PR #87, opened 2025-07-20
Add Docker configurations and helper script for development and production environments

nipuna-perera:
Closes #85 , Fixes #86 

This pull request introduces significant updates to the docker workflow of SparkyFitness application, including detailed documentation, improved Docker configurations, and enhanced development workflows. The changes focus on streamlining deployment, improving environment management, and providing comprehensive guidance for developers. Below are the most important changes grouped by theme:

### Documentation Enhancements
* Added a detailed architecture and development workflow guide in `.github/copilot-instructions.md`, covering the tech stack, repository patterns, AI integration, and database schema conventions.
* Updated `README.md` to include clear instructions for production and development setups, along with a new section on using the Docker helper script.

### Docker Configuration Improvements
* Introduced a new `docker-helper.sh` script for managing Docker environments, supporting commands like `up`, `down`, `build`, and `clean`.
* Enhanced `docker-compose.dev.yml` for development with live reloading and exposed ports, and updated `docker-compose.prod.yml` for production with pre-built images. [[1]](diffhunk://#diff-740e0cdeda40047aeb493482ba2b352edbacdc0107a84491d149bcd219dd95ffR1-R62) [[2]](diffhunk://#diff-a46123669fb686fe41f9f8c23d2cb510b5ced2f0636b86eba4e455d460fc0d84L1-L2)
* Added a frontend entrypoint script (`docker/docker-entrypoint.sh`) to dynamically configure Nginx using environment variables, ensuring flexibility in deployments.

### Codebase and Build Process Updates
* Updated `Dockerfile.backend` and `Dockerfile.frontend` to streamline multi-stage builds, including the addition of `gettext` for environment variable substitution in the frontend. [[1]](diffhunk://#diff-fac4519b568b1338edc2406be290153c56be62a01b857dead95e48d9de5de428L27-R27) [[2]](diffhunk://#diff-2f48856a5903f2f6f401dea10d3b0375d433c5132a4802a7627f86eb3aaf255bL18-R42)
* Reorganized Docker-related files, including the renaming of `docker-compose.yml` to `docker-compose.prod.yml` and the introduction of `nginx.conf.template`. [[1]](diffhunk://#diff-a46123669fb686fe41f9f8c23d2cb510b5ced2f0636b86eba4e455d460fc0d84L12-R10) [[2]](diffhunk://#diff-da8fcbe728a9172b578e5d754f8e2df214c658c4321f610e63dd68bea828ab49R1-R142)

### Development Workflow Improvements
[...]

PR #88, opened 2025-07-20
Minor docs fix

jay-oswald:
* Fixed project name on docs title, forgot to change it earlier
* Add links to docs to readme
* Removed homepage from docs nav


jay-oswald:
These got added to #87

PR #90, opened 2025-07-20
Docker workflow build

jay-oswald:
The first commit adds in building and publishing to ghcr, 2nd commit I added in publishing to dockerhub.

I've tested deploying to ghcr and its all working: https://github.com/jay-oswald/SparkyFitness/pkgs/container/sparkyfitness-frontend

I've tweaked the names slightly for ghcr to standerdise it on the GH standard naming, dockerhub uses the existing names exacily.

This gets triggered every time a new release is made on GH, so deploy process is just add the commits, create the release, and let the workflow build and deploy.

GH auth is auto-magic, but for dockerhub will need to add 2 secrets to the repository: `DOCKER_USERNAME` and `DOCKER_PASSWORD`

### DIFF
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -17,11 +17,11 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v4
 
-      # - name: Log in to Docker Hub
-      #   uses: docker/login-action@v3
-      #   with:
-      #     username: ${{ secrets.DOCKER_USERNAME }}
-      #     password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
 
       - name: Log in to the Container registry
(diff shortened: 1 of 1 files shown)


nipuna-perera (commented):
Is it necessary to push to docker hub? I think just ghcr.io is good enough. Most OSS projects just push to ghcr now. It's free and there are no rate limiting shenanigans.

.github/workflows/docker-deploy.yml
```diff
@@ -0,0 +1,86 @@
...
+        id: meta-frontend
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            codewithcj/sparkyfitness
+            ghcr.io/${{ github.repository }}-frontend
+
+      - name: Build and push frontend
+        id: push-frontend
+        uses: docker/build-push-action@v6
```
nipuna-perera, line 46: Does this do multi-arch natively? If not that should be considered here.

jay-oswald: That part happens in the build step, already setting it to build for both AMD and ARM

nipuna-perera: Ah I see it now. I'm not very experienced with GHA.

CodeWithCJ (approved):
thank you!!!

PR #120, opened 2025-08-08
Feature: Include brand when searching

gtkirk:
## Description
Concatenate brand and name when searching to allow pattern searching over both columns.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Testing
- [x] Changes have been tested locally
- [x] No console errors or warnings
- [x] Mobile responsiveness verified (for UI changes)

## Screenshots (if applicable)
image

image

### DIFF
--- a/SparkyFitnessServer/models/foodRepository.js
+++ b/SparkyFitnessServer/models/foodRepository.js
@@ -38,10 +38,10 @@ async function searchFoods(name, userId, exactMatch, broadMatch, checkCustom) {
     let paramIndex = 1;
  
     if (exactMatch) {
-      query += `f.name ILIKE $${paramIndex++} AND f.user_id = $${paramIndex++}`;
+      query += `CONCAT(f.brand, ' ', f.name) ILIKE $${paramIndex++} AND f.user_id = $${paramIndex++}`;
       queryParams.push(name, userId);
     } else if (broadMatch) {
-      query += `f.name ILIKE $${paramIndex++} AND (f.user_id = $${paramIndex++} OR f.is_custom = FALSE)`;
+      query += `CONCAT(f.brand, ' ', f.name) ILIKE $${paramIndex++} AND (f.user_id = $${paramIndex++} OR f.is_custom = FALSE)`;
       queryParams.push(`%${name}%`, userId);
     } else if (checkCustom) {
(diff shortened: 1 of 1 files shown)


CodeWithCJ (changes requested):
Is there a way to let users know this feature exist directly in the UI? May be a help icon or foot note with example patterns. Otherwise no one will be aware this exits. And I am afraid people doesn't have patience to read through 1000s of lines of doc even if we setup doc for this in future. what do you think? 
Thanks for your contributions.

gtkirk:
I probably overcomplicated the examples; the '%' is a wildcard supported by SQL's LIKE/ILIKE that matches any sequence of zero or more characters. This change just concatenates the food brand and name with a space so they both can be searched.

So you can search and find results that match the brand:
image

and you can also find the brand followed by the first word(s) in the food name without any special patterns:
image

It might be useful to mention SQL's built in pattern matching functionality, but I was also wondering if you were open to some additional changes to search that could supersede that. Splitting the query on whitespace and chaining multiple ILIKE clauses with AND could take the place of '%' and also make word order unimportant. PostgreSQL also has some interesting advanced [text search](https://www.postgresql.org/docs/current/functions-textsearch.html) options like supporting and, or, and not operators, though this may be overkill.

Let me know if you'd like me to explore any of that on a new PR.

CodeWithCJ:
Does it work for foodname space brandname as well?  There are multiple places the search are available such as diary, food, meal, meal plan etc. could you review if this change is consistent in all those places as well. 
What you have should be fine for now.  Based on future user request, we can further fine tune. but now with space between the words, should be enough for now as far as it works both ways i.e. foodname space brandname and as well as brandname space foodname.

gtkirk:
No, word order still matters because it's still using the ILIKE operator, e.g. searching for '%chocolate%cookie%' would find 'chocolate chip cookie', but not 'cookie with chocolate chips'. Finding results with any word order would require splitting the query up into multiple clauses ANDed together, or using another type of pattern matching.

I can confirm the changes work on the diary, food manager, meal editor, and meal plan editor.

PR #159, opened 2025-09-21
Fixes for 0% calorie distribution goals

gtkirk:
Various fixes for calorie goal setting and display:
* Use nullish coalescing operator (??) instead of logical or (||) when setting default goals so that 0 values are not overridden
* Update conditional rendering in MealCard.tsx to better handle 0 calorie meal goals
* Pass changedMeal down to normalizePercentages so that it is not considered when reallocating percentages

Before:
Breakfast slider/input can't be set below 1 without using Intermittent Fasting setting:
image

0% calorie goals show up as 25% of calories on diary (showing 0%/0%/50%/50% distribution):
image

After:
Breakfast slider/input can be set to 0:
image

0% calorie goals are not shown (showing 0%/0%/50%/50% distribution):
image


CodeWithCJ (approved):

CodeWithCJ:
@gtkirk Thanks a lot for all these improvements. Please connect with me via discord if you are there. I will add you to developer channel.

PR #189, opened 2025-10-02
Update healthConnectService.js

KraLeoD:
Fixed metrics incorrectly using record.startTime instead of record.time


KraLeoD:
This should enable the Android App to pick up metrics which the app isn't fetching from health connect right now

CodeWithCJ (approved):

PR #193, opened 2025-10-03
fix MFA for Garmin and add Garmin China by env:IS_CN

IamTaoChen:


IamTaoChen:
Maybe, adding a flag named 'is_china' on the web is better

CodeWithCJ:
@IamTaoChen  Thanks for your contribution. Could you clarify the following.

- Is the code change only for China or can it be re-used for users from any country that are enabled with MFA? If this is common for anyone, could you rename the the env variable accoridnly? 
- what is the use of REQUESTS_CA_BUNDLE in the env variable? Is it indirectly used anywhere?

IamTaoChen:
> * Is the code change only for China or can it be re-used for users from any country that are enabled with MFA? If this is common for anyone, could you rename the the env variable accoridnly?

I only have a Chinese account at the moment, so I have not tested it with other countries.



> * what is the use of REQUESTS_CA_BUNDLE in the env variable? Is it indirectly used anywhere?

I use `REQUESTS_CA_BUNDLE` specifically because https://sso.garmin.cn was causing SSL errors. That’s the reason I added it.

CodeWithCJ:
Thanks for the clarification. Probably change that to generic name e.g. is_mfa_active = true then use MFA function. I can test from end.. This way it doesn't be one country specific.

PR #213, opened 2025-10-14
Refactor display of nutrient values into NutrientGrid

gtkirk:
When searching for foods from the diary, Recent and Top Foods display nutrients in a single column despite the intention to use additional columns:
image

This is because the sm:grid-cols-5 class name is constructed dynamically: https://github.com/CodeWithCJ/SparkyFitness/blob/20109b259983909f25267688711ea9163cdb0c45/SparkyFitnessFrontend/src/components/EnhancedFoodSearch.tsx#L760  and Tailwind is not generating the style for it. Tailwind discourages this: https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names

This PR adds a utility class to select css class names similar to https://github.com/CodeWithCJ/SparkyFitness/blob/20109b259983909f25267688711ea9163cdb0c45/SparkyFitnessFrontend/src/pages/Index.tsx#L349 and encapsulates the display of nutrient values into a reusable NutrientGrid component.

image

### DIFF
--- a/SparkyFitnessFrontend/src/components/EnhancedFoodSearch.tsx
+++ b/SparkyFitnessFrontend/src/components/EnhancedFoodSearch.tsx
@@ -70,6 +70,70 @@ interface EnhancedFoodSearchProps {
 
 type FoodDataForBackend = Omit<CSVData, "id">;
 
+const NutrientGrid = ({ food, visibleNutrients }) => {
+  const nutrientDetails: { [key: string]: { label: string; unit: string } } = {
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
Similar conflict issue as I updated the App. Please re-submit the PR for this change as well. Appreciate your help on this,

gtkirk:
I think this should work, unless FatSecret was providing some non-falsy value for glycemic_index that had to be overridden explicitly with "None"

PR #220, opened 2025-10-16
Fix share buttons for Foods and Exercises pages

gtkirk:
Previously the web page would hang when trying to share a food. This was a result of the togglePublicSharing click handler shadowing the togglePublicSharing foodService method it intended to call, resulting in infinite recursion. Resolved by renaming the click handler.

Previously sharing an exercise also didn't work. Recent changes to the exerciseRouter expect form data containing exerciseData and associated images instead of exerciseData on its own.
image

Resolved by updating the request and also ensuring images will only be updated if they were part of the request (otherwise they could be deleted inappropriately when changing the sharing status).

### DIFF
--- a/SparkyFitnessFrontend/src/services/exerciseService.ts
+++ b/SparkyFitnessFrontend/src/services/exerciseService.ts
@@ -149,9 +149,12 @@ export const deleteExercise = async (id: string, userId: string): Promise<void> 
 };
 
 export const updateExerciseShareStatus = async (id: string, sharedWithPublic: boolean): Promise<Exercise> => {
+    const payload = new FormData();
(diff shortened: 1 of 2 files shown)


CodeWithCJ (approved):

CodeWithCJ:
@gtkirk Though the exercise and food appear in other profile, still they couldn't be used in diary or other sections. When I implemented this originally in Supabse, I had RLS. But with local Postgres lot of things changed in the way server interacts with DB and Frontend. 

- There are few scenarios like what will happen if the shared food is revoked but its used by others. Others who use will lose access to it suddenly. Same scenario for for Exercise as well
- How to share Meals and Workout presets. Because they continue numerous food & exercise. 

If you are in discord, please message me to chat further on this.

Issue #238, opened 2025-10-21
[Bug]: Backup/Restore issues


gtkirk:
## Bug Description

Making and restoring backups runs into a few issues:

1. Manual backup requests hang after the database dump. It appears pgDump closes after awaiting the pipeline to gzip/output so there is no close or error event to resolve the next promise: https://github.com/CodeWithCJ/SparkyFitness/blob/9224265d4f0e4ffb9d7209aef384ad560f7549d4/SparkyFitnessServer/services/backupService.js#L79

Commenting out the whole promise seems to allow the process to complete and create a full backup, so this will probably need to be reworked to add the event handlers earlier or explicitly check if the process has already finished.

2. When the backup folder is mapped to a separate volume, attempting to restore a backup results in the following error:
logging.js:30 [2025-10-21T01:31:33.364Z] [ERROR] Error during restore: Error: EXDEV: cross-device link not permitted, rename '/app/SparkyFitnessServer/temp_uploads/ea789261753b31d23e04d01bd4ad76ed' -> '/app/SparkyFitnessServer/backup/sparkyfitness_full_backup_2025-10-21T01-27-56-157Z.tar.gz'

Looks like this may need to be broken up into a copy/unlink or fs-extra.move.

3. After restoring a backup, repeated errors are thrown about using a pool after calling end:
logging.js:30 [2025-10-21T02:03:55.968Z] [ERROR] Error caught by centralized handler: Cannot use a pool after calling end on the pool Error: Cannot use a pool after calling end on the pool

Looks like this is coming from a stale pool reference in the session middleware. There is an attempt to reconfigure the middleware with the new pool:
https://github.com/CodeWithCJ/SparkyFitness/blob/9224265d4f0e4ffb9d7209aef384ad560f7549d4/SparkyFitnessServer/services/backupService.js#L199
Which is never called because the app parameter isn't supplied when calling performRestore:
https://github.com/CodeWithCJ/SparkyFitness/blob/9224265d4f0e4ffb9d7209aef384ad560f7549d4/SparkyFitnessServer/services/backupService.js#L199

When it is called, however, it only adds a new session middleware to the end of express' stack, which still leaves the older instance higher up in the stack with the stale pool reference. Will probably require [express-dynamic-middleware](https://github.com/lanbomo/express-dynamic-middleware#readme) or a similar approach to replace the session middleware at the appropriate place in the stack.

## Steps to Reproduce

1. As the administrator, backup and restore Sparky Fitness

[...]

PR #240, opened 2025-10-21
Fixes for backup/restore

gtkirk:
Fixes issues in #238 

- wrap pipeline/pgDump promises in Promise.all to avoid contention when awaiting
- split file rename into copyFile/unlink to avoid potential cross-device link not permitted error
- refactor configureSessionMiddleware to only modify sessionMiddleware and not alter the middleware stack
- add indirection layer when use-ing sessionMiddleware so the new version will be called if it is replaced
- ensure performRestore calls configureSessionMiddleware after resetting postgres pool

### DIFF
--- a/SparkyFitnessServer/SparkyFitnessServer.js
+++ b/SparkyFitnessServer/SparkyFitnessServer.js
@@ -165,14 +165,9 @@ app.get(
 
 let sessionMiddleware; // Declare sessionMiddleware globally
 
-const configureSessionMiddleware = (app, pool) => {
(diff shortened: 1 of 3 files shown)


CodeWithCJ:
I changed many things in the server and front end side as I am trying to setup DB RLS to properly implement sharing foods & exercise. Please re-submit the PR after couple of days as I am continue to make changes. The code version I have is currently unstable. So wait for couple of days before you update yours. Thanks for understanding!!!

PR #280, opened 2025-10-30
Onboarding Backend SQL fix

shantnudon:
- There was an error in the migration SQL that was resulting in problems in setting up the on-boarding.


CodeWithCJ:
You shouldn't change the existing migrating script as for those it was already executed, it will not be applied. Add as new script with latest timestamp.

shantnudon:
The script must have ran for all the users who have the latest version or even they are new. When we start the server the all the migrations are done. I have attached a screenshot where i have applied to an existing table.

image

PR #293, opened 2025-11-03
Fixed onboarding SQL

shantnudon:
The onboarding SQL has been fixed.
Commit 1 has the file name changed
Commit 2 has the SQL changes that makes sure that the TABLES are created only if they dont exists.


CodeWithCJ:
@shantnudon  This will not be executed as the table already exits from migration file. you need the ALTER the table if you are changing the structure and as well as replace the trigger.

shantnudon:
I am not altering the table nor the trigger. I m just making sure that if they are there then let them be else create them.

PR #323, opened 2025-11-12
Fix checkin mood duplicate & allow '0' or 'tired' to be valid mood values for /api/mood

the-fifth:
Noticed the 'check-in' mood slider has two entries for 'tired' emoji, as it counts at 0 on the slider and also at 10.
image
image

Modified the mood api route to allow 0 as a valid value, this was actually a bug on the current live unrelated to the fact there was a duplicate 'tired' mood option, you are unable to submit a mood value of 0.

I'm not proficient in React/Typescript so if there's something amiss with the changes please feel free to disregard or give feedback on what needs changing.

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -56,8 +56,10 @@ server {
   }
 
 
-  location /uploads/ {
-    proxy_pass http://${SPARKY_FITNESS_SERVER_HOST}:${SPARKY_FITNESS_SERVER_PORT};
+  # Proxy requests for uploaded files to the backend.
+  # Using a regex match to ensure higher precedence and correct routing.
+  location ~ ^/uploads/(.*)$ {
+    proxy_pass http://${SPARKY_FITNESS_SERVER_HOST}:${SPARKY_FITNESS_SERVER_PORT}/uploads/$1;
     proxy_set_header Host $host;
     proxy_set_header X-Real-IP $remote_addr;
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
I think removing 0 from the mood meter scale might be the fix needed. I saw you changed Neutral from 50 to 40. Now what it is going to do is those who previously saved their day as Neutral are now going to be shown as  Thoughtful which would be wrong.

So I would suggest to leave getMoodDisplay as before but remove 0 from the scale to it doesn't show tired twice and probably start the scale from 10. 

Let me know if any question.

the-fifth:
> I think removing 0 from the mood meter scale might be the fix needed. I saw you changed Neutral from 50 to 40. Now what it is going to do is those who previously saved their day as Neutral are now going to be shown as Thoughtful which would be wrong.
> 
> So I would suggest to leave getMoodDisplay as before but remove 0 from the scale to it doesn't show tired twice and probably start the scale from 10.
> 
> Let me know if any question.

Thanks for getting back to me, I reverted the slider range values to make sure the previously entered moods would reflect and simply clamped the slider to a minimum of 10, it should eliminate the issue with both submitting a 0 mood value and removing the duplicate.

CodeWithCJ:
I noticed the comment section is not aligned with the actual code. it doesn't have any effect. So I will take care of updating it alter.

image

PR #344, opened 2025-11-16
remove missed call to refresh settings on manual backup

the-fifth:
in relation to pull #341, I missed that there was a call to refresh the backup settings upon running a manual backup. Not sure how settings could/would change upon that and if they would affect the manual backup process. Moving it into the useEffect(...) hook in the previous pull makes it obviously unreachable from outside of it.
Seeing as this is just a deletion/removal of two lines I'm sure easy enough to throw into an ongoing change, no need to specifically merge this one.

### DIFF
--- a/SparkyFitnessFrontend/src/pages/Admin/BackupSettings.tsx
+++ b/SparkyFitnessFrontend/src/pages/Admin/BackupSettings.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { api } from '../../services/api'; // Assuming an API service exists
 import { useAuth } from '../../hooks/useAuth'; // Import useAuth hook
@@ -21,8 +21,7 @@ const BackupSettings: React.FC = () => {
   const daysOfWeek = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'];
 
   // Fetch current backup settings and status from backend
-  useEffect(() => {
-    const fetchBackupSettings = async () => {
+  const fetchBackupSettings = async () => {
     try {
       const response = await api.get('/admin/backup/settings');
       const data = response || {}; // Ensure data is an object even if response is null/undefined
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
Sorry not clear. Did you now fix the manual backup refresh status as well or that is something I will need to look into it?

CodeWithCJ:
@gtkirk  would know better as he made some fixes on this. 

https://github.com/CodeWithCJ/SparkyFitness/pull/240

the-fifth:
The PR was to remove the call to refresh the backup settings during/after a manual backup. If still desire to do that, I could move the function back out of useEffect and simply call it within the hook, I just wanted to clarify if there was any point to refreshing the backup settings immediately after a manual backup? They're inherently not related, ie why do you need to know automatic backup settings when explicitly running a manual one.

gtkirk:
The backup settings also contain the status and timestamp for the last backup, so it does need to call fetchBackupSettings after a manual backup to update those values.

In #341 was it really necessary to move the definition of fetchBackupSettings into the anonymous function supplied to useEffect? Or was removing it from the array of dependencies enough to fix the constant querying?

the-fifth:
> In #341 was it really necessary to move the definition of fetchBackupSettings into the anonymous function supplied to useEffect? Or was removing it from the array of dependencies enough to fix the constant querying?

It was a solution I found while looking into how it worked/functions, I have compiled and tested locally that you are (obviously) correct in that just removing it from the depency array allows it to stop querying. Apologies, my inexperience with React caused me to overcomplicate the solution.

Reverting to commit `dce61fb` and removing from the dependency array should achieve the desired result without disrupting the functionality, how did you want to proceed?

CodeWithCJ:
> Apologies, my inexperience with React caused

No need of apology!!! I appreciate your help on this and thanks a lot for helping me in making this App useful for everyone!!!

the-fifth:
I have restored the file to `dce61fb` as noted, removed it from the dependency array. I will stick to raising issues rather than opening PRs until I am more proficient, thanks both!

CodeWithCJ:
Thanks for your help. We are all learning. So don't worry. Feel free to submit PRs for improvements and dont let this small issue stop you!!!

CodeWithCJ (approved):

Issue #346, opened 2025-11-17
[Bug]: SPARKY_FITNESS_API_ENCRYPTION_KEY lenght


DrDimedrol:
Trying to setup using docker compose. Can't run because get error 
"[ERROR] Encryption key (SPARKY_FITNESS_API_ENCRYPTION_KEY) has an invalid length. Expected 64 hex characters or 44 Base64 characters. Update your environment variable and try again."
Trying to fill .env file variable with "openssl rand -base64 64" and "44" command output, but nothing helps.

the-fifth:
Are you running it from the `./docker/docker-helper.sh` script? I think it might be expecting the `.env` file to be in the same directory as the compose file.

Temporary solution:
Move your `.env` file to be in the `docker/` directory where the `.env.example` file currently is, that will allow the docker compose process to detect it in the directory. If you've correctly followed the directions and used `openssl rand -hex 32` it should actually pick it up.

CodeWithCJ:
@DrDimedrol  Let us know if you are still having issue. you can also google to generate 64 hex code and use in SparkyFitness in case if you can't run the command given by  @the-fifth  .  

https://numbergenerator.org/random-64-digit-hex-codes-generator

CodeWithCJ:
Closing as there are no response. Reach out to us again if any issues

PR #347, opened 2025-11-17
Explicitly pass --env-file flag to docker compose command to load .env from project directory in alignment with docs

the-fifth:
Currently the docs instruct user to `cp docker/.env.example .env` during setup and update the values, and the `docker-helper.sh` script fails to find the `SPARKY_FITNESS_API_ENCRYPTION_KEY` and `JWT_SECRET` and throws errors on container up (#346)

I think the change from `docker-compose` to `docker compose` changed the behaviour with regards to how it looks for `.env` files, `docker-compose` would (according to old github issues and google searches at least) look in parent/working directory until it found one, whereas `docker compose` will only check the directory of the compose file itself by default.

(I know at least with this I can confirm I had to modify my `docker-helper.sh` locally to get it up and running)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ __pycache__/
 SparkyFitnessServer/uploads/
 SparkyFitnessServer/backup/
 SparkyFitnessServer/temp_uploads/
+
+# SparkyFitnessGarmin specific ignores
+SparkyFitnessGarmin/mock_data/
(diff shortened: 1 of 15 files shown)


CodeWithCJ:
docker-helper.sh was added by one of the community user. Personally I think its too many place to maintain when there is a new change to compose or env file . So I am going to remove it and it will be easier to just directly use the compose and env file.

PR #349, opened 2025-11-17
Added Slovenian translation

thehijacker:
I started working on Slovenian translation but since I have tools to translate faster than I can do it on Weblate GUI, I have translated the json file directly. Find it in this PR. Can you please add it? I will update the future needed translation later on Weblate:

https://hosted.weblate.org/projects/sparkyfitness/-/sl/

But for first push please use this translation.

Thank you.


CodeWithCJ:
@thehijacker  Could you check on this "GitGuardian has uncovered 1 secret following the scan of your pull request."
Not sure if its confused with the translation. I will wait for you to check before I merge this!!

thehijacker:
> @thehijacker Could you check on this "GitGuardian has uncovered 1 secret following the scan of your pull request." Not sure if its confused with the translation. I will wait for you to check before I merge this!!

I have no idea where that warning is comming from. There are no hardcoded secrets in the file. I marked to solve, but it only added new line at end. I am sure this is some sort of false positive. It is safe to merge.

PR #449, opened 2025-12-30
Fix Garmin sync for body composition metrics

Alexvbp:
## Summary
My Mi Fit scale syncs body composition metrics (body water percentage, muscle mass, visceral fat level) to Garmin Connect. When syncing Garmin data to SparkyFitness, these metrics failed with database errors because they were mapped as `check_in` table columns that don't exist.

Changed these three metrics to use custom measurements instead (like BMI and Bone Mass already do).

## Changes
- `garminMeasurementMapping.js`: Changed `body_water_percentage`, `muscle_mass`, and `visceral_fat_level` from `check_in` to `custom` type
- `measurementService.js`: Removed these fields from the check-in switch case (they now fall through to custom measurement handling)

## Test plan
- [x] Sync Garmin data with a scale that reports body water, muscle mass, or visceral fat
- [x] Verify no database column errors in logs
- [x] Verify measurements appear in custom measurements


CodeWithCJ:
Thank you for your contribution. It will help everyone.

PR #451, opened 2025-12-30
Fix AI chatbot bubble visibility when no active AI provider

Alexvbp:
## Summary
The AI chatbot bubble remained visible and unresponsive when all AI providers were disabled. Clicking it did nothing because there was no active service to connect to.

## Root Cause
Two issues:
1. The component was calling `getAIServices()` which returns all services (including disabled ones) instead of `getActiveAiServiceSetting()` which only returns active services
2. When no active service exists, the API returns an empty object `{}` rather than `null`. Since `{} !== null` is `true`, the bubble remained visible

## Changes
- **DraggableChatbotButton.tsx**: 
  - Switched from `getAIServices()` to `getActiveAiServiceSetting()`
  - Added check for empty objects: `Object.keys(activeService).length > 0`
  - Added visibility change listener to re-check when returning to the tab
  - Added periodic polling (every 5 seconds) to catch changes without refresh

## Test plan
- [X] Add an AI provider and verify the bubble appears
- [X] Disable the AI provider and verify the bubble disappears within 5 seconds
- [X] Re-enable the AI provider and verify the bubble reappears
- [X] Refresh the page with no active provider and verify bubble is hidden

### DIFF
--- a/SparkyFitnessFrontend/src/components/DraggableChatbotButton.tsx
+++ b/SparkyFitnessFrontend/src/components/DraggableChatbotButton.tsx
@@ -1,40 +1,115 @@
 import React, { useState, useEffect, useRef, useCallback } from 'react';
-import { Button } from "@/components/ui/button";
-import { MessageSquare } from 'lucide-react';
+import { X } from 'lucide-react';
 import { useChatbotVisibility } from '@/contexts/ChatbotVisibilityContext';
 import { useIsMobile } from '@/hooks/use-mobile';
 import { getActiveAiServiceSetting } from '@/services/aiServiceSettingsService';
 import { useAuth } from '@/hooks/useAuth';
 
+const BUTTON_SIZE = 56; // 14 * 4 = 56px (w-14)
+const MINIMIZED_SIZE = 24;
+const EDGE_MARGIN = 16;
+const BOTTOM_NAV_HEIGHT = 80; // Height of bottom navigation bar on mobile
+const STORAGE_KEY = 'chatbot_button_state';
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
Actually it was bot icon instead of bubble. But somehow I broke it and it turned to be bubble :).
I couldn't fix it to to be better visual after multiple tries. So i left it with bubble. If you could fix, it would be good.
I was trying with both below gif and small video. But you can also intorudce better visual for the bot icon.

SparkyFitnessFrontend\public\images\chatbot.gif

SparkyFitnessFrontend\public\videos\chatbot.mp4

Alexvbp:
## Summary
- Replaced the SVG bubble with the `chatbot.gif` image
- Added full drag-to-reposition functionality that works on both desktop and mobile
- Button now snaps to the nearest edge (left or right) when released
- Added minimize/dismiss feature: hover X button on desktop, drag to edge on mobile
- Button respects mobile bottom navigation bar and won't overlap it
- Smooth animations for entrance, dragging, and restoring from minimized state

## Test plan
- [X] Verify chatbot.gif displays correctly
- [X] Test dragging the button on desktop (mouse)
- [X] Test dragging the button on mobile (touch via DevTools)
- [X] Verify button snaps to nearest edge when released
- [X] Test minimize via X button hover on desktop
- [X] Test minimize by dragging to edge on mobile
- [X] Verify minimized pill can be tapped to restore
- [X] Check button doesn't overlap bottom nav on mobile viewport
- [X] Disable AI provider and verify button disappears
- [X] Force reload page and verify button is still draggable

PR #455, opened 2026-01-01
fixes sticky zero values

### LINKED ISSUE #333: [Feature]:   Sticky Zero Values in Fields
In numeric fields, a zero is pre-populated.  This zero is sticky and needs to be removed manually. Consider displaying a placeholder zero instead which would disappear when the field received focus.  You could then just default any field not touched to zero on data writes.

Alexvbp:
Fixes #333 

## Summary

- Nutrient input fields in food entry forms now show empty with a grayed-out "0" placeholder instead of a pre-populated sticky zero
- Users can click and immediately type values without manually deleting the zero first
- Empty fields are treated as 0 when saving to the database

### CustomFoodForm.tsx
- Updated `CustomFood` interface to allow `number | ""` for all nutrient fields
- Changed initial state from `0` to `""` for all 18 nutrient fields
- Updated `handleInputChange` to preserve empty strings for nutrient fields
- Added `placeholder="0"` to all nutrient Input components
- Convert empty strings to 0 on form submission

### EnhancedCustomFoodForm.tsx
- Added `FormFoodVariant` type allowing `number | ""` for nutrient fields
- Added `formVariantToFoodVariant()` helper (converts "" → 0 for saving)
- Added `foodVariantToFormVariant()` helper (converts 0 → "" for display)
- Updated state from `FoodVariant[]` to `FormFoodVariant[]`
- Updated all variant initializations to use "" instead of 0
- Added `placeholder="0"` to all nutrient Input components
- Convert form variants before saving via `formVariantToFoodVariant()`

## Test plan

- [X] Open CustomFoodForm and verify nutrient fields are empty with "0" placeholder
- [X] Click a nutrient field and type a value - should work without deleting zero
- [X] Leave some fields empty and save - verify they save as 0
- [X] Open EnhancedCustomFoodForm and verify same behavior
- [X] Edit an existing food with 0 values - verify fields show as empty
- [X] Verify custom nutrients also show placeholder behavior

### DIFF
--- a/SparkyFitnessFrontend/src/components/DraggableChatbotButton.tsx
+++ b/SparkyFitnessFrontend/src/components/DraggableChatbotButton.tsx
@@ -1,220 +1,388 @@
 import React, { useState, useEffect, useRef, useCallback } from 'react';
-import { Button } from "@/components/ui/button";
-import { MessageSquare } from 'lucide-react';
+import { X } from 'lucide-react';
 import { useChatbotVisibility } from '@/contexts/ChatbotVisibilityContext';
 import { useIsMobile } from '@/hooks/use-mobile';
-import { getAIServices } from '@/services/aiServiceSettingsService';
+import { getActiveAiServiceSetting } from '@/services/aiServiceSettingsService';
 import { useAuth } from '@/hooks/useAuth';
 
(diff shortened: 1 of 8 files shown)


CodeWithCJ:
Thank you for your contribution. if would like to join the dedicated developer channel in discord, please connect with me in discord.

Issue #357, opened 2025-11-20
[Bug]: Garmin periodic sync fails


thehijacker:
## Important: Log Requirement for Troubleshooting

```
[2025-11-20T05:00:00.033Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T06:00:00.023Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T07:00:00.021Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T08:00:00.022Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T09:00:00.045Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T10:00:00.023Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T11:00:00.023Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T12:00:00.039Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T13:00:00.037Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
[2025-11-20T14:00:00.032Z] [ERROR] Error during scheduled Garmin data sync: garminConnectService.getGarminDailySummary is not a function
```

## Bug Description

It seems periodic Garmin sync is not working. I can see errors in the server log.

## Steps to Reproduce

-

## Expected Behavior

Garrmin activities and health data should sync.

## Actual Behavior

It doesn't sync.

## Environment

- **SparkyFitness Version:**  0.15.9.3
- **Running Env:** Docker

CodeWithCJ:
Did you try this with 0.15.9.4 ?

thehijacker:
I was on 0.15.9.4. with manual docker image tag to get to that version. Now using latest.  I am not sure it is doing periodic garmin checks now. Nothing it logs.

CodeWithCJ:
Check the modified timestamp for mock_data to confirm. If it has latest timestamp then background refresh is working fine

[...]

PR #465, opened 2026-01-03
Garmin historical sync, UI enhancements and code adjustments

Alexvbp:
## Overview

This PR adds the ability to sync multiple years of historical Garmin data, replacing the previous 7-day limit. Users can now import their complete Garmin history with full progress visibility.

##  What's New

Historical Import with Date Range Selection

  - New "Historical Import" button in Garmin provider settings
  - Date picker dialog to select custom start/end dates
  - Preset buttons for common ranges (1 year, 2 years, 5 years)
  - Option to skip already-synced data for efficient re-runs

Large date ranges are split into 7-day chunks processed sequentially. This approach:
  - Prevents timeouts on long-running requests
  - Respects possible Garmin API rate limits
  - Add health checks to docker compose

 ## Technical Implementation

Backend:
  - New garmin_sync_jobs table tracks job state, progress, and failed chunks
  - Job service handles chunking, processing, and state management
  - Repository layer with proper transaction handling and RLS support

Frontend:
  - HistoricalImportDialog for date range selection
  - GarminSyncProgress component with real-time updates
  - Polling mechanism that reliably picks up new jobs
  - Success state display on completion

Infrastructure:
  - New database migration for sync jobs table
  - CLI script for running imports outside the UI

  ## Code Quality

  - Race condition prevention with database-level locking
  - SQL field allowlist to prevent injection
  - JSDoc documentation on all public functions
  - Refactored CLI to use service layer (no logic duplication)
  - Proper error handling and logging throughout

  ## Additional Fixes

  - Reports table: JSON data (like Raw Stress Data) now shows collapsed with "X data points" summary instead of rendering massive unformatted blobs

## Testing

- [X] Connect a Garmin account
- [X] Click "Historical Import"
- [X] Select a date range (try "Last 2 Years" preset)
- [X] Verify progress updates smoothly
- [X] Let it complete and verify success message
- [X] Check sleep and training data

Possible fix for #357

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -12,6 +12,12 @@ services:
       - sparkyfitness-network
     ports:
       - "5432:5432" # Expose database port for local development
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${SPARKY_FITNESS_DB_USER:-sparky} -d ${SPARKY_FITNESS_DB_NAME:-sparkyfitness_db}"]
+      interval: 5s
+      timeout: 5s
+      retries: 10
+      start_period: 10s
 
   sparkyfitness-server:
     build:
@@ -51,7 +57,14 @@ services:
       - "3010:3010" # Expose backend port for local development
     restart: always
     depends_on:
-      - sparkyfitness-db
+      sparkyfitness-db:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:3010/health"]
+      interval: 5s
+      timeout: 5s
+      retries: 10
+      start_period: 30s
     volumes:
       - ../SparkyFitnessServer:/app/SparkyFitnessServer # Mount source for live reloading
       - /app/SparkyFitnessServer/node_modules # Preserve node_modules
@@ -70,8 +83,10 @@ services:
       - sparkyfitness-network
     restart: always
     depends_on:
-      - sparkyfitness-server
-      - sparkyfitness-garmin
+      sparkyfitness-server:
+        condition: service_healthy
+      sparkyfitness-garmin:
+        condition: service_started
 
   sparkyfitness-garmin:
     build:


CodeWithCJ (approved):

CodeWithCJ (changes requested):
Currently, exercise activities are stored as JSON in exercise_entry_activity_details. This works well because it allows custom fields to be created dynamically when a user logs an exercise from the web.

However, for Garmin integration, it would be better to store this data in a structured way using proper tables and fields. Doing so would require significant changes across the server, database, frontend, and the Garmin integration itself.

That said, this approach would be beneficial in the long run. If we design the table and field structure to be generic, it could support integrations not only with Garmin, but also with Fitbit, Apple, and Android exercise/workout data.

It would require an understanding of data models across all these platforms, but we could start with Garmin and incrementally extend support to the others over time.

What are your thoughts on this approach?

Alexvbp:
Yeah, I think that's the right call. JSONB makes sense for manual exercise logging where fields can vary, but for platform integrations the data is consistent enough that proper tables are cleaner.
I'll close this PR to work on the activity schema restructure, afterwards i can look at the enhancements.

Today i played around with the following:

  Database Schema:
  - activities - Normalized core metrics (duration, distance, HR, speed, etc.)
  - activity_laps / activity_heart_rate_zones - Detailed breakdown data
  - activity_samples - Time-series GPS/HR/speed with PostGIS support
  - activity_type_mappings - Normalizes provider-specific types (e.g., Garmin "trail_running" → "running/trail")
  - platform_data JSONB - Preserves provider-specific extras we don't normalize

  Key Decisions:
  - Backwards compatible - falls back to old JSONB for legacy entries
  - Auto-migration on server startup for upgrading users
  - Generic design ready for multiple providers

  Frontend:
  - Activity reports use new normalized API
  - Added speed/HR color-coded map visualization

  Status:
  - Garmin working but requires more testing
  - Next: Fitbit and HealthKit to validate the generic design

I'll continue testing and tinkering for a while, it is quite a significant change like you mentioned.

PR #473, opened 2026-01-04
Create add-missing-translations script

Korri:
Add a script that identifies missing keys for a specific language and adds them to the translation file for that language with a `[TODO] ` prefix

This will help when new keys get added so they can easily be found and added by translators.

### DIFF
--- a/SparkyFitnessFrontend/package.json
+++ b/SparkyFitnessFrontend/package.json
@@ -10,7 +10,8 @@
     "build:dev": "vite build --mode development",
     "lint": "eslint .",
     "preview": "vite preview",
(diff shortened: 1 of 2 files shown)


CodeWithCJ:
Hi @Korri  Thanks for your contribution. I am having trouble understanding this PR's usage. Weblate already shows what are the missing translations. Is this really needed? Now by adding ToDO and filling dummy translation weblate may not know it needs to be updated isnt it?

Korri:
> Weblate already shows what are the missing translations.

🫣 I wasn't aware that this project used Weblate, ignore this then!

PR #476, opened 2026-01-04
Update translations for exercises in German

mf450s:


CodeWithCJ:
Could you update this via https://hosted.weblate.org/engage/sparkyfitness/  

To avoid out of sync, I would prefer updating via Weblate

Issue #468, opened 2026-01-03
[Feature]: Serving Size for Food Import - Use Per Serving Instead of Per 100g


Gtt1229:
## Is your feature request related to a problem? Please describe.

I am running into what I think is an issue of sorts. During import -> Edit and Add, there is no way to properly use Unit Type because it is tied to the 100G calculation from OpenFoodFacts.

For example, I can not accurate depict 1 serving, since it defaults to the 100G serving size, rather than the serving size itself.

## Describe the solution you'd prefer

I propose that the 100g measurement (energy-kcal_100g)should be used for grams, while other arbitrary, non-standardized measurements should default to the serving measurement, energy-kcal_serving

## Describe alternatives you've considered
N/A

## Additional context
I realized this may be better as a Feature Request. Initial discussion post here:

https://github.com/CodeWithCJ/SparkyFitness/discussions/467

BryceKrispiess:
I believe that’s just how openfoodfacts works, if you look at their website, it’s always per 100g.

Gtt1229:
> I believe that’s just how openfoodfacts works, if you look at their website, it’s always per 100g.

Possibly for poorly entered, but I listed the JSON key that is used in the API request that returns the calories per serving, energy_kcal_serving. Serving is arbitrary by nature though.

Gtt1229:
To expand on this, we have the following via json for an ingredient that has a serving size populated rather than 100g:

| Field | Value |
|------|-------|
| energy-kcal_serving | 120 |
| serving_quantity | 64 |
| serving_quantity_unit | g |
| energy-kcal_100g | 188

The serving quanity is a value per serving in grams, this is 1 serving and is a computed value in grams. The serving quantity unit is irrelevant for this calculation.

We would effectively have to calculate the ratio between serving quanity vs energy-kcal_100g. In this case, we have a 0.64, so all values need to be multiplied by .64 to get their actual value per serving.

This would apply to arbitrary measurements like "piece, slice, etc" otherwise, what is really the point of their existence in conjunction with Open Food Facts?

Basically, most food products are not 100 grams per serving. Open Meal Facts has a natural limitation due to the (correct) decision to still to proper units. This can be remediated.

Gtt1229:
I found this thread which kind of touched on it and has a workaround using the autoscaling feature: https://github.com/CodeWithCJ/SparkyFitness/issues/89

[...]

PR #501, opened 2026-01-09
Add option to autoscale foods via OpenFoodFacts

Gtt1229:
This is my take on my feature request:

https://github.com/CodeWithCJ/SparkyFitness/issues/468

Creates an option (disabled by default) that can be toggled that will enable scaling of serving size automatically form OpenFoodFacts

I am not too confident in my implementation for best practices of this project as a whole, so please forgive me.

### DIFF
--- a/SparkyFitnessFrontend/src/components/EnhancedFoodSearch.tsx
+++ b/SparkyFitnessFrontend/src/components/EnhancedFoodSearch.tsx
@@ -1338,49 +1338,56 @@ const EnhancedFoodSearch = ({
 
         {activeTab === "online" &&
           openFoodFactsResults.length > 0 &&
-          openFoodFactsResults.map((product) => (
-            <Card
-              key={product.code}
-              className="hover:bg-gray-50 dark:hover:bg-gray-700"
-            >
-              <CardContent className="p-4">
-                <div className="flex justify-between items-start">
-                  <div className="flex-1">
-                    <div className="flex items-center space-x-2 mb-2">
-                      <h3 className="font-medium">{product.product_name}</h3>
-                      {product.brands && (
-                        <Badge variant="secondary" className="text-xs">
-                          {product.brands.split(",")[0]}
+          openFoodFactsResults.map((product) => {
+            // Calculate display values based on auto-scaling preference
+            const shouldScale = autoScaleOpenFoodFactsImports && product.serving_quantity && product.serving_quantity > 0;
+            const servingSize = shouldScale ? product.serving_quantity! : 100;
+            const scaleFactor = shouldScale ? servingSize / 100 : 1;
+            
+            return (
+              <Card
+                key={product.code}
+                className="hover:bg-gray-50 dark:hover:bg-gray-700"
+              >
+                <CardContent className="p-4">
+                  <div className="flex justify-between items-start">
+                    <div className="flex-1">
+                      <div className="flex items-center space-x-2 mb-2">
+                        <h3 className="font-medium">{product.product_name}</h3>
+                        {product.brands && (
+                          <Badge variant="secondary" className="text-xs">
+                            {product.brands.split(",")[0]}
+                          </Badge>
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
Could you attach screenshots to understand better on this.

Gtt1229:
- Adds the following setting:

image

Example food stats:

| Nutrition facts | As soldfor 100 g / 100 ml | As soldper serving (1 patty (64 g)) |
| --- | ---- | ---- |
| Energy | 784 kj(188 kcal) | 502 kj(120 kcal) |

- **THIS IS UNCHANGED!** During Import (Normal display of Search and Add Food remains):
image

- After clicking "Edit and Add" the scaling of 1 serving is applied during import, rather than having to scale, then adjust:

image

It utilizes the serving size to perform a scaling factor to apply to the nutriments. If it is disabled or the serving size is already 100g, the import is unaffected/not-scaled.

CodeWithCJ (approved):

CodeWithCJ:
is anything pending or shall i go ahead & merge it? where you able to test that it doesn't affect other food providers?

Gtt1229:
I unfortunately don't have any other providers to test at the moment. I can set one up some time today; with that, the scaling takes place in the `convertOpenFoodFactsToFood` constant, so I don't foresee it affecting other imports.

CodeWithCJ:
Hi @Gtt1229 there was a big change from another PR. So I had to hold this one as the other one involved changing lot of pages. Now your PR has some conflict. could you re-submit again. sorry for the inconvenience. Thanks for your help..

Gtt1229:
> Hi @Gtt1229 there was a big change from another PR. So I had to hold this one as the other one involved changing lot of pages. Now your PR has some conflict. could you re-submit again. sorry for the inconvenience. Thanks for your help..

I made the changes, just looks like the lists were cleaned up. I will test my instance later today, please hold on merge.

Gtt1229:
I added identical logic to the search results for OpenFoodFacts as well for alignment:

image


Tested and ready to go.

frozendevl:
@Gtt1229  @CodeWithCJ  Just wanted to say THANK YOU for adding auto scaling, incredibly useful and solves one of my biggest gripes with adding new food. Do I need to check "auto-scale" on the Add Custom Food screen? I've been clicking it because it seems to not work otherwise, but wanted to see if it's user error somewhere. Also, is it possible to do the same for "FatSecret" provider? I've had better luck using them (larger database and faster results).

Gtt1229:
> @Gtt1229 Just wanted to say THANK YOU for adding auto scaling, incredibly useful and solves one of my biggest gripes with adding new food. Do I need to check "auto-scale" on the Add Custom Food screen? I've been clicking it because it seems to not work otherwise, but wanted to see if it's user error somewhere. Also, is it possible to do the same for "FatSecret" provider? I've had better luck using them (larger database and faster results).

Of course!

And its 2 different things. Under Settings -> Preferences -> You have to enable and save: Auto-scale OpenFoodFacts Imports.

The treatment following is the results from OoenFoodFacts will be scaled to the serving size during import.

Auto scale can be used in the event you need to change during import (or edits in general).

Basically, I generally do 2 entries for my imports. 1 entry for the grams, and then another duplicate entry as the default for "prices/bar/serving/etc". If you don't do the grams, you can't reliably do measurement if you need it. I'll upload some images of the whole process later today to better specify.

CodeWithCJ:
> I'll upload some images of the whole process later today to better specify.

If possible, submit as PR to improve the doc on this part..

Gtt1229:
@frozendevl 

Here is the documentation breaking it all down: https://codewithcj.github.io/SparkyFitness/features/settings/preferences

I'll take a look at FatSecret. Never used it.

Gtt1229:
@frozendevl Based in their API docs, it looks to me like FatSecret should already return results based on their servings sizes.

Issue #516, opened 2026-01-14
[Bug]: Health data is synced to the next day from Android companion app


Gtt1229:
## Bug Description

When syncing data from the mobile companion app on Android (may apply to IOS), the data for today is synced to tomorrow.

## Expected Behavior

Data sync to the correct day.

Gtt1229:
The AI did a lot of work for not much reason.

The logic that is currently deployed is as follows:

App:
```
return new Date(date as string | number | Date).toISOString().split('T')[0];
```

Server:
```
  const entriesToClean = healthDataArray.filter(
    d => d.type === 'SleepSession' || d.type === 'ExerciseSession' || d.type === 'Workout'
  );

  if (entriesToClean.length > 0) {
    const datesBySource = {};

    for (const entry of entriesToClean) {
      const source = entry.source || 'manual';
      // Replicate date parsing logic from main loop to ensure consistency
      const dateToParse = entry.date || entry.entry_date || entry.timestamp;
      if (dateToParse) {
        const dateObj = new Date(dateToParse);
        if (!isNaN(dateObj.getTime())) {
          const parsedDate = dateObj.toISOString().split('T')[0];
          if (!datesBySource[source]) {
            datesBySource[source] = {};
          }
          // use object keys for unique dates set
          datesBySource[source][parsedDate] = true;
        }
      }
    }
```

The timestamps are being converted from UTC prior to being sent to the server, so the mobile app is already attempting to parse, when the timestamp should actually be sent to the server as UTC. I am working on improving the logic, rather than just processing the timestamp via the mobile app.

Gtt1229:
Upon further review, the database doesn't store entry date in UTC, which is why it has to be accurate from the mobile app prior to hitting the server.

PR #519, opened 2026-01-15
Mobile App Time Zone Fix

Gtt1229:
This is in reference to https://github.com/CodeWithCJ/SparkyFitness/issues/516.

**I do not have an iPhone that I am able to test on.**

Previously, when using the mobile app, I would receive data for the next rather than same day. The time stamp was originally parsed via: `const date = timeToUse.split('T')[0];`. This formats the timestamp in UTC and then splits the yyyy-mm-dd, without the offset.

The new logic gets the startTime and digests the string/date, then uses the TS Date interface to convert a UTC string to locale, then formats the output to match yyyy-mm-dd.

```
export const toLocalDateString = (timestamp: string | Date): string => {
  const localDate = new Date(timestamp);
  return `${localDate.getFullYear()}-${String(localDate.getMonth() + 1).padStart(2, '0')}-${String(localDate.getDate()).padStart(2, '0')}`;
};
----
const date = toLocalDateString(record.startTime);

```

Server-side logs you can see the log in UTC, while the data sent is in the correct timezone, EST.

```
sparkyfitness-server-1    | [2026-01-15T00:37:51.481Z] [INFO] Incoming health data JSON: [
sparkyfitness-server-1    |   {
sparkyfitness-server-1    |     "value": 718,
sparkyfitness-server-1    |     "type": "step",
sparkyfitness-server-1    |     "date": "2026-01-14",
sparkyfitness-server-1    |     "unit": "count"
sparkyfitness-server-1    |   }
sparkyfitness-server-1    | ]
```

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -12,67 +12,57 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
-
+    
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
-
+      
     - name: Set up JDK 17
       uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
         java-version: '17'
-
+        
     - name: Set up Node.js
       uses: actions/setup-node@v4
       with:
         node-version: '20'
-        cache: 'npm'
-        cache-dependency-path: SparkyFitnessMobile/package-lock.json
-
-    - name: Configure Gradle properties
-      run: |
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
@apedley  Could you review this

apedley:
Looks good. Time zones was on my list of things to check out, thanks @Gtt1229 !

CodeWithCJ:
Thank you @apedley  @Gtt1229

Issue #492, opened 2026-01-08
[Bug]: Garmin activity minutes are off by a factor of 60


deusrex2k:
## Bug Description

When an activity from Garmin synchs, Sparky fitness logs too many minutes by a factor of 60. 

For example, for a 50 minute Garmin workout, Sparky lists 3000 minutes.

## Steps to Reproduce

1. Complete a workout in Garmin
2. Synch the activity to Sparky Fitness 
3. Sparky fitness logs the time as actual minutes * 60

## Expected Behavior

Duration matches between Garmin and Sparky fitness

## Environment

- **SparkyFitness Version 0.16.3.2
- **Running Env: docker

CodeWithCJ:
could you provide screenshot from garmin connect and as well as from sparkyfitness.

deusrex2k:
![Image](https://github.com/user-attachments/assets/10048818-a3d5-4226-b2b8-eba579a3915c)
![Image](https://github.com/user-attachments/assets/d25f13ed-aee1-4c9e-9d49-0a4f46961e28)
![Image](https://github.com/user-attachments/assets/a1216ed7-727e-4e77-b817-ab4df0916752)

The issue appears on the Home Page, the Exercise page looks OK.

CodeWithCJ:
Fix will be available in next release.

PR #603, opened 2026-02-06
fix (garmin): #492 garmin activity minutes off by factor 60

Sim-sat:
## Description

Fixes: #492 

The activity minutes are off by factor 60. Minutes are expected but seconds are passed. The division by 60 is missing here.

Before:
grafik

After:
grafik

Connect
Image 2026-02-06 at 11 07 49


Sim-sat:
> Using Math.round() here may lead to a loss of precision for set durations. For instance, a 29-second set would be rounded to 0 minutes, while a 30-second set would become 1 minute. Since the total duration_minutes for the exercise entry on line 342 is not rounded, it would be more consistent and accurate to also store the individual set duration as a floating-point number. This preserves more granular data.

It's not possible since a whole number is expected. Not using Math.round() leads to errors

CodeWithCJ (approved):

PR #608, opened 2026-02-07
fix: default .env download path in docs

StephenFluin:
The file seems to have been moved in recent releases.


StephenFluin:
Don't merge this. The path is right but the curl command still didn't work.

CodeWithCJ:
I will have to rename the env file to match the one from docker folder as well. github workflow needs some updates as well.

CodeWithCJ:
I was able to fix the curl command.

PR #621, opened 2026-02-08
fix(health-data): apply default units for HealthConnect sync (fixes #567)

### LINKED ISSUE #567: [Bug]: HealthConnect Default Units
## Bug Description

When pulling my health data from healthconnect units are not carrying over so I am seeing N/A. Shouldn't there be default units used for that data?

## Steps to Reproduce

1. Use an app with health connect support(gadgetbridge for me).
2. Sync data using the SparkyFitness app
3. Go to the web ui
4. You see everything besides steps and sleep with N/A next to it.

## Expected Behavior

All data should have units set correctly so that the graphs work and data is correct.

## Actual Behavior

All imported data besides steps seems to not have a unit, so thus graphs are wrong and data format.

## Screenshots/Videos (Optional)

Image

JosephDoUrden:
## Summary
Fixes #567: HealthConnect (and other client) sync was creating custom measurement categories with `measurement_type = 'N/A'` because the backend did not use the `unit` field from the payload and had no defaults. The web UI then showed "N/A" for those metrics and graphs were wrong.

## Changes
- **Backend:** In `processHealthData`, when handling scalar health data as custom measurements, the category's `measurement_type` is now taken from `dataEntry.unit` (or `dataEntry.measurementType`) when present; otherwise a default is applied from a new `DEFAULT_UNITS_BY_HEALTH_TYPE` map (e.g. heart_rate → bpm, TotalCaloriesBurned → kcal, distance → m), aligned with the mobile HealthMetrics.
- **Tests:** Added `measurementService.healthDataUnits.test.js` to assert default units when unit is missing and that payload unit is used when provided.

## Testing
- `npx jest tests/measurementService.healthDataUnits.test.js` (with `SPARKY_FITNESS_API_ENCRYPTION_KEY` set) — 4 tests pass.
- Manual: sync HealthConnect data for heart rate / calories / distance and confirm the web UI shows correct units instead of N/A.

## Notes
- Existing categories that were already created with `measurement_type = 'N/A'` are unchanged; users can edit the category in Settings to set the unit, or re-sync after deleting those entries if desired. New syncs will get the correct default.

### DIFF
--- a/SparkyFitnessServer/services/measurementService.js
+++ b/SparkyFitnessServer/services/measurementService.js
@@ -310,9 +310,12 @@ async function processHealthData(healthDataArray, userId, actingUserId) {
           // Handle as custom measurement
           // Use unit from payload (e.g. HealthConnect sends "unit") or default so UI does not show "N/A"
           const unitFromPayload = dataEntry.unit ?? dataEntry.measurementType;
-          const resolvedMeasurementType = (unitFromPayload && String(unitFromPayload).trim() !== '')
-            ? String(unitFromPayload).trim()
-            : (DEFAULT_UNITS_BY_HEALTH_TYPE[type] || 'N/A');
+          let resolvedMeasurementType;
+          if (unitFromPayload && typeof unitFromPayload === 'string' && unitFromPayload.trim()) {
+            resolvedMeasurementType = unitFromPayload.trim();
+          } else {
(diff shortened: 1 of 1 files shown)


CodeWithCJ (approved):

JosephDoUrden:
Thanks, happy to take another issue @CodeWithCJ

PR #626, opened 2026-02-08
chore: use eslint to remove unused imports

Sim-sat:
## Description

Used eslint plugin to remove all unused imports


CodeWithCJ:
@Sim-sat  @asmello  why is the CI test failing? Should CI test needs any improvement or its issue with the code?

asmello:
This was an issue with testing, but seems that the fix had been added already in #626, and this branch was just out of sync with main.

Timeline:
  1. PR #626's head commit (3212a9f) has parent 72d9ec1 (the merge of PR #624 "Fix mobile view") — that's the point the branch was based on.
  2. PR #624 introduced the useIsMobile hook using window.matchMedia, breaking tests.
  3. The fix commit 3d5ee1e ("fix (test): add matchMedia to test env") landed on main after that — it's 1 commit ahead of 72d9ec1 with 0 commits behind. So PR #626 was branched from the exact point where tests were broken, before the fix arrived.

asmello:
Note CI checks were failing in https://github.com/CodeWithCJ/SparkyFitness/actions/runs/21799572758/job/62892739672, and it was merged anyway. Maybe we should start enforcing checks to pass, but allow you to bypass? This should make it less likely that failures will be missed.

CodeWithCJ:
Yes. I ignored the error and merged by bypass option. But it would be good to fix the CI test so that it can be utilized effectively. BUt it might have already been fixed based on other PR and as you also confirmed the same.

asmello:
> But it would be good to fix the CI test so that it can be utilized effectively.

There was nothing wrong with the CI testing itself, it's just that if someone changes how things work tests will sometimes need to be updated. The issue was in #624 itself, it should've not been merged as it was.

Sim-sat:
> Note CI checks were failing in https://github.com/CodeWithCJ/SparkyFitness/actions/runs/21799572758/job/62892739672, and it was merged anyway. Maybe we should start enforcing checks to pass, but allow you to bypass? This should make it less likely that failures will be missed.

Yes I know. I had the fix ready but it was merged before I could push it. 
100%, Checks should be enforced. We should use eslint to enforce the same formatting as well.

Sim-sat:
@asmello  we should also add tsc -b to the build command like a normal ts project. there are 29 errors that have to be resolved before tho

CodeWithCJ:
@Sim-sat  @asmello  Thank you both!!!

PR #640, opened 2026-02-09
Fix cache (like) problems

Sim-sat:
## Description

There were issues with the new version that only got resolved after manually deleting the website data and cache. The reason is the vite pwa worker that was set to prompt, but is never prompted. I set the mode to autoUpdate so it updates every time.
In the process I improved the nginx config to exclude the /assets folder from the regex folder. Only assets inside this folder should be cached for 1y because of the hash value vite adds to the filename.

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -69,10 +69,9 @@ server {
     proxy_set_header X-Forwarded-Ssl on; # Explicitly tell backend that connection is SSL
   }
    location /assets/ {
-    root /usr/share/nginx/html;
     expires 1y;
     add_header Cache-Control "public, no-transform, immutable";
-    try_files $uri =404; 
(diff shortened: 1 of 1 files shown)


Sim-sat:
@CodeWithCJ can't we just remove the pwa worker? The only benefit is being able to use the app in offline mode, but every actions needs the server anyway. I think it's not useful for this use case

CodeWithCJ:
We need PWA as it helps to view the web in full screen like native app in mobile. So many requested for this to be PWA

CodeWithCJ (approved):

Sim-sat:
> We need PWA as it helps to view the web in full screen like native app in mobile. So many requested for this to be PWA 

You're right I forgot it's needed for that.

Issue #172, opened 2025-09-28
[Feature]: Adding support for Polar flow


marmata:
## Is your feature request related to a problem? Please describe.

Since there are integrations for Garmin and other ecosystems, lets have Polar Flow as well, as probably the second biggest sport watch vendor

## Describe the solution you'd prefer

Seamless integration with Polar Flow as it’s now for Garmin

## Describe alternatives you've considered

Manual inputting of data is one alternative.

gerryhendrickx:
I second this! It would be awesome if we can feed polar data into Sparky automatically to have everything in one place!

CodeWithCJ:
@marmata  @gerryhendrickx @nielsvdp  integration for polar flow is completed.  But note the following.

- PolarFlow doesnt allow history sync it seems after you established connection with SparkyFitness, only the data added after will be sync. Not sure if my understanding correct, but as per API doc this is what they are saying.
- So I tested with onyl few sample and the manual exercise data I added in their web is not sync. Not sure if this another limitation or some bug in code. 

Set SPARKY_FITNESS_SAVE_MOCK_DATA=true to your Server container, Sync PolarFlow to generate JSON under mock_data. share me the anonymized file to me via Discord if you would like to contribute further enhancements on this.

PR #663, opened 2026-02-15
Integration with PolarFlow

CodeWithCJ:
https://github.com/CodeWithCJ/SparkyFitness/issues/172


Sim-sat:
@CodeWithCJ  I suggest merging #661 first because it makes the code simpler and I already created an extra component for the screen with the message that can be used.

CodeWithCJ:
@Sim-sat  I am not good at resolving code conflicts. Also as your is small change, do you think its easier to fix again yours than this PR ?

Sim-sat:
@CodeWithCJ There are no conflicts it's just that that the use of a direct api call and no using the component are "step back" from my pr. Not saying they are wrong, but history wise I'd make more sense to first merge the correct solution and then add something on top. The effort in changing it is very small though

CodeWithCJ:
@Sim-sat  merged all PRs. could you have a look at Polarflow and make tweaks if needed similar to what you did for other callbacks

Sim-sat:
@CodeWithCJ yes I will. You can look at them to see that it's not much different and how it makes the code easier to understand. Instead of calling the api function directly you wrap it in a useQuery hook that is just copy and paste from the other ones. 

Also I would not blindly use AI code. For example:  
`navigate('/?tab=settings&section=integrations'); 
` makes absolutely no sense. We didn't even have routes before I migrated to tanstack router and even now there's only a /settings route.
Can you make sure to format the code using prettier before committing? That would prevent the changes in the code that are only because of different line endings. If you use VSCode you can just do that in settings I think. Or you can run npx prettier . --write before commiting (even set it as a pre commit hook to do automatically) or use npm run format.

CodeWithCJ:
> navigate('/?tab=settings&section=integrations');

yes, i noticed that as well. But left it for now.

PR #670, opened 2026-02-15
integration with hevy

CodeWithCJ:


Sim-sat:
@CodeWithCJ  the connectHeavyAccount function is not used

PR #715, opened 2026-02-22
Dev QoL on Fedora fixes

toddejohnson:
I use podman more than docker and it is to be drop in compatible. To use the [Quick start](https://codewithcj.github.io/SparkyFitness/developer/getting-started#quick-start-docker-recommended) I needed these fixes:

* Pass `-env-file "$(pwd)/.env" ` on down too
*  Add `:z` to volumes so SELinux doesn't get angry

I don't think these should affect anyone on other distributions. It should make it easier for the next Fedora dev.


CodeWithCJ:
@Sim-sat  could you review this once

Sim-sat:
Looks good. Even fixes an issue I had today in debian.

PR #956, opened 2026-03-21
remove unwanted code from last pr

Sim-sat:
I  tested a view all exercise history page but I think the diary is the wrong place for it and forgot to delete the link to it.


CodeWithCJ:
@Sim-sat will you be fixing the type script error?

Sim-sat:
Didnt see it. Just forgot to remove the import

PR #957, opened 2026-03-21
Fix: Garmin duplicate activities on same day overwrite each other instead of creating separate entries

Soulplayer:
## Description

Two Garmin activities of the same type on the same day (e.g. two cycling rides) caused only one entry to survive sync. The second activity silently overwrote the first instead of creating a new row.

**Root cause:** `garminService.js` never passed Garmin's `activityId` as `source_id` when creating exercise entries. The dedup logic in `exerciseEntry.js` therefore fell back to matching on `(exercise_id, entry_date, source)` — which is identical for two cycling activities on the same day — and updated the existing row instead of inserting a new one.

## Fix

- `garminService.js`: populate `source_id` with `activity.activityId` in both `processGarminSimpleActivity` and `processGarminWorkoutSession`
- `exerciseEntry.js`: when `source_id` is present and no match is found by source_id, skip the manual `(exercise_id + date)` fallback — a miss on source_id means it's a genuinely different activity that must INSERT

## Behaviour after fix

| Scenario | Before | After |
|---|---|---|
| Two cycling rides on the same day | Only one row — second overwrote first | Two separate rows ✅ |
| Same activity re-synced (same `activityId`) | Updated existing row | Still updates (idempotent) ✅ |
| Manual entries (no `source_id`) | Normal dedup | Unchanged ✅ |
| Strava / Polar / Withings | Normal dedup | Unaffected ✅ |

### DIFF
--- a/SparkyFitnessServer/services/garminService.js
+++ b/SparkyFitnessServer/services/garminService.js
@@ -453,7 +453,7 @@ async function processGarminSimpleActivity(userId, activityData) {
     notes: `Garmin Activity: ${activity.activityName} (${activity.activityType?.typeKey})`,
     distance: activity.distance,
     avg_heart_rate: activity.averageHR || activity.averageHeartRateInBeatsPerMinute || null,
-    source_id: activity.activityId?.toString() || null,
+    source_id: activity.activityId?.toString() ?? null,
   };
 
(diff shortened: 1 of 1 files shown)


CodeWithCJ:
@Soulplayer I committed the gemini feedback directly to this PR.

PR #1021, opened 2026-03-28
feat: data source priority settings per health category

Soulplayer:
## Description

Adds a new **Data Source Priority** panel to the Settings page that lets users choose  
which integration to use as the primary source for each health data category. When  
multiple integrations report data for the same day, the selected source is preferred  
over the automatic quality ranking introduced in the companion measurement dedup PR.

Supported categories:
- **Sleep** — duration, stages, HRV, SpO₂  
- **Body & Weight** — weight, BMI, body fat, muscle mass  
- **Daily Activity** — steps, distance, active calories  
- **Workouts** — always merged from all sources (no preference needed; duplicates  
  are handled by the cross-source exercise dedup PR)

### Changes

**Database** (single migration file: `20260328110000_add_source_preferences_to_user_preferences.sql`)
- Adds `sleep_source_preference`, `body_source_preference`, `activity_source_preference`  
  columns to `user_preferences` (default `'auto'`)

**Backend**
- `preferenceRepository`: `getUserPreferences()` and `saveUserPreferences()` now  
  read/write the three new source preference columns
- `measurementService`: `getCheckInMeasurements()` loads the user's body and activity  
  preferences and passes them to `findPreferredCheckInRow()`, which returns the  
  preferred source's row instead of always returning the quality-ranked first row
- `sleepRoutes`: passes the user's sleep source preference into the sleep data retrieval  
  so the preferred source is returned when multiple sources exist for a night

**Frontend**
- `PreferencesContext`: exposes `sleepSourcePreference`, `bodySourcePreference`,  
  `activitySourcePreference` and `saveAllPreferences()` to the rest of the app
- `DataSourceSettings`: new accordion panel in Settings showing three dropdowns  
  (Sleep / Body & Weight / Daily Activity) plus a read-only Workouts row explaining  
  the automatic merge behaviour; includes a Save button
- `SettingsPage`: registers the new `DataSourceSettings` accordion item

## Related Issue

PR type [x] New Feature  
Linked Issue: N/A

## Checklist

- [ ] **[MANDATORY for new feature] Alignment**: Data source priority per category is a  
  natural extension of the multi-source deduplication work; no blocking design decisions.
- [ ] **Tests**: UI preferences are persisted via existing preference infrastructure;  
  server-side preference loading is covered by the measurement dedup test suite (PR 1).
- [x] **[MANDATORY for UI changes] Screenshots**: See Before / After below.
- [x] **[MANDATORY for Frontend changes] Quality**: `pnpm run validate` — typecheck ✅  
  lint ✅. Pre-existing prettier issues in upstream unrelated to this PR.
- [x] **Translations**: No translation strings added or changed.
- [x] **Architecture**: Follows existing `PreferencesContext` pattern; new component  
  placed alongside other settings components.
- [x] **Database Security**: `user_preferences` is already covered by RLS in  
  `rls_policies.sql`; the three new columns do not change access control.
- [x] **[MANDATORY - ALL] Integrity & License**: I certify this is my own work, free of  
  malicious code and I agree to the License terms.

### Before

Settings page had no section for choosing preferred data source per category.
Schermafbeelding 2026-03-28 190238

### After

![Data Source Priority settings panel showing Sleep, Body & Weight, Daily Activity dropdowns and Workouts explanation]
_(New "Data Source Priority" accordion in Settings — users can pin a specific  
integration per category or leave it on Auto for quality-ranked automatic selection)_
Schermafbeelding 2026-03-28 165355

### DIFF
--- a/db_schema_backup.sql
+++ b/db_schema_backup.sql
@@ -2296,6 +2296,9 @@ CREATE TABLE public.user_preferences (
     default_barcode_provider_id uuid,
     auto_scale_online_imports boolean DEFAULT true,
     first_day_of_week smallint DEFAULT 0,
+    sleep_source_preference character varying(50) DEFAULT 'auto'::character varying NOT NULL,
+    body_source_preference character varying(50) DEFAULT 'auto'::character varying NOT NULL,
+    activity_source_preference character varying(50) DEFAULT 'auto'::character varying NOT NULL,
     CONSTRAINT check_energy_unit CHECK (((energy_unit)::text = ANY (ARRAY[('kcal'::character varying)::text, ('kJ'::character varying)::text]))),
     CONSTRAINT logging_level_check CHECK ((logging_level = ANY (ARRAY['DEBUG'::text, 'INFO'::text, 'WARN'::text, 'ERROR'::text, 'SILENT'::text])))
 );
@@ -2306,6 +2309,27 @@ CREATE TABLE public.user_preferences (
 --
 
 COMMENT ON COLUMN public.user_preferences.auto_scale_open_food_facts_imports IS 'When enabled, OpenFoodFacts imports will automatically scale nutrition values from per-100g to the serving size provided by the product';
+
+
+--
+-- Name: COLUMN user_preferences.sleep_source_preference; Type: COMMENT; Schema: public; Owner: -
+--
+
+COMMENT ON COLUMN public.user_preferences.sleep_source_preference IS 'Preferred source for sleep analytics: auto | garmin | withings | fitbit | polar | healthkit | health_connect | manual';
+
+
+--
+-- Name: COLUMN user_preferences.body_source_preference; Type: COMMENT; Schema: public; Owner: -
+--
+
+COMMENT ON COLUMN public.user_preferences.body_source_preference IS 'Preferred source for body & weight data: auto | withings | garmin | fitbit | healthkit | health_connect | manual';
+
+
+--
+-- Name: COLUMN user_preferences.activity_source_preference; Type: COMMENT; Schema: public; Owner: -
+--
+
+COMMENT ON COLUMN public.user_preferences.activity_source_preference IS 'Preferred source for daily activity (steps, distance, calories): auto | garmin | fitbit | polar | healthkit | health_connect | manual';
 
 
 --


CodeWithCJ:
@Sim-sat  @apedley  Could you review this.

Sim-sat:
Mmh. I don't think I would hardcode the providers. Wouldn't a single field or multiple fields be enough to have a priority? So having a drag and drop menu to order them and only saving the order as numbers? Just an Idea though. 
The problem I see is that every Addition of a provider has a new place where it has to be added even If we have endpoints for the current providers

Soulplayer:
Yeah I had the same thought but I bookmarked it for later. It would definetly be better then how it is now done.

Sim-sat:
> Yeah I had the same thought but I bookmarked it for later. It would definetly be better then how it is now done. 

Yes, but wouldn't that be simpler to implement as well? I'd prefer to do it right because nobody will probably touch it again when it's working

SparkyFitnessServer/models/measurementRepository.js
```diff
@@ -227,51 +228,65 @@ async function upsertCheckInMeasurements(
...
-      return null; // Return null if no measurements to update/insert
+      return null;
     }
 
-    const existingRecord = await client.query(
-      'SELECT * FROM check_in_measurements WHERE user_id = $1 AND entry_date = $2',
-      [userId, entryDate]
+    // Source-aware upsert: each source keeps its own row per date.
+    // ON CONFLICT relies on the unique constraint (user_id, entry_date, source)
+    // added in migration 20260328100000.
```
apedley, line 236: where is that migration?

SparkyFitnessFrontend/src/pages/Settings/DataSourceSettings.tsx
```diff
@@ -0,0 +1,221 @@
...
+  };
+
+  return (
+    <AccordionItem value="data-sources" className="border rounded-lg mb-4">
+      <AccordionTrigger
+        className="flex items-center gap-2 p-4 hover:no-underline"
+        description="Choose which integration to use as the primary source per data category"
+      >
+        <Database className="h-5 w-5" />
+        Data Source Priority
```
apedley: Data Source is pretty vague. Maybe something about Integration Priority

SparkyFitnessServer/db/migrations/20260328110000_add_source_preferences_to_user_preferences.sql
```diff
@@ -0,0 +1,17 @@
+-- Migration: Add data source preferences to user_preferences
+-- Users can now select a preferred data source per category (sleep, body, activity).
+-- Default 'auto' means the existing score-based best-source selection is used.
+
+ALTER TABLE public.user_preferences
```
apedley, line 5: If you add any columns you need to update the schema file `db_schema_backup.sql` in the root directory.

SparkyFitnessServer/models/measurementRepository.js
Soulplayer: Ow it is not included. Thank you for reporting this!

Soulplayer:
I'm gonna close this because without Garmin these implementations are not necessarily better then the app.

PR #1091, opened 2026-04-07
fix: include source in exercise_snapshot for v2 by-date endpoint

johnpena:
## Summary
- The v2 `_buildExerciseEntryWithSnapshot` function omits `source` from the `exercise_snapshot` object
- The frontend checks `exercise_snapshot.source` to decide whether to prepend `/uploads/exercises/` to image paths
- Without `source` in the snapshot, it's `undefined` (falsy), so the frontend uses raw image filenames as URLs, resulting in 404s for all exercise images in the diary view

## Test plan
- [ ] Navigate to the diary page with exercise entries that have images from FreeExerciseDB
- [ ] Verify exercise images load correctly (no broken image icons)
- [ ] Verify the image URLs include the `/uploads/exercises/` prefix

### DIFF
--- a/SparkyFitnessServer/services/exerciseEntryHistoryService.ts
+++ b/SparkyFitnessServer/services/exerciseEntryHistoryService.ts
@@ -116,6 +116,7 @@ function _buildExerciseEntryWithSnapshot(
       id: entryData.exercise_id as string,
       name: exercise_name as string,
       category: (category as string) ?? null,
+      source: (source as string) ?? null,
       images: _parseJsonArray(images),
       primary_muscles: _parseJsonArray(primary_muscles),
       secondary_muscles: _parseJsonArray(secondary_muscles),


apedley:
5 lines above, source_id is already sent so I don't know if sending the data twice is the right fix. I don't see where the frontend would expect it in the snapshot object. I only see it referenced like:

```
exerciseEntry.source
	? `/uploads/exercises/${snapshot.images[0]}`
	: snapshot.images[0]
```


Are you seeing image 404s?  Where?

johnpena:
Thanks for the review! You're right that the latest source in `ExerciseEntryDisplay.tsx` checks `exerciseEntry.source` (top level), which the v2 endpoint already provides. My deployed frontend was on an older version that checked `exercise_snapshot.source` instead, which is why I was seeing 404s on exercise images in the diary view.

I'll update my deployment to the latest frontend, which should resolve it without this change. Happy to close this if you think including `source` in the snapshot is unnecessary — though it could be argued it belongs there for completeness since the snapshot schema already defines it as an optional field.

CodeWithCJ:
Thanks for your contribution.  This is no longer needed.

johnpena:
Quick correction on my previous reply: I checked v0.16.5.6 (the current release tag), and `ExerciseEntryDisplay.tsx` there still uses `exerciseEntry.exercise_snapshot.source`, not `exerciseEntry.source`. The change to use the top-level `source` only exists on `main` and hasn't been released yet.

So this fix is needed for users on v0.16.5.6 (which is the version distributed via the Proxmox community script). Once a release goes out with the frontend change, this PR becomes redundant and can be closed. Happy to defer to the maintainer on whether to merge this as a stopgap or just wait for the next release.

apedley:
A release is planned for later today and everything should work then. Let me know if you still have issues

PR #1257, opened 2026-05-10
Fix(strava): round float average_heartrate before insert (#1256)

### LINKED ISSUE #1256: [Bug]: Strava sync fails for all activities — invalid input syntax for type integer (avg_heart_rate float)
## Version

v0.16.6.1 (Docker `codewithcj/sparkyfitness_server:v0.16.6.1`, Postgres 18.3-alpine)

## Bug

Manual Strava sync (`POST /api/integrations/strava/sync`) responds with `{"success":true,"source":"live_api","activitiesProcessed":4}` but **0 entries are actually written to `exercise_entries`** because every activity throws a Postgres type error.

## Server log

```
[ERROR] [stravaDataProcessor] Error processing activity "Morning Run": invalid input syntax for type integer: "126.9"
    at async _createExerciseEntryWithClient (/app/SparkyFitnessServer/models/exerciseEntry.ts:497:27)
    at async Object.createExerciseEntry (/app/SparkyFitnessServer/models/exerciseEntry.ts:579:19)
    at async Object.processStravaActivities (/app/SparkyFitnessServer/integrations/strava/stravaDataProcessor.ts:166:24)
    at async Object.syncStravaData (/app/SparkyFitnessServer/services/stravaService.ts:184:7)
```

## Root cause

`exercise_entries.avg_heart_rate` is declared as `integer` in schema, but the Strava API returns `average_heartrate` as a float (e.g. `126.9`). In `integrations/strava/stravaDataProcessor.ts:151` the value is passed through unchanged:

```ts
avg_heart_rate: activity.average_heartrate || null,
```

Postgres rejects the insert.

## Side effect

`activitiesProcessed` counter in `stravaService.ts` is incremented per loop iteration regardless of success — so the API returns success and a non-zero count even when every insert fails. Sync looks healthy from the outside while no data lands.

## Suggested fix

Two changes (either alone fixes the symptom; combining them is more robust):

1. Round in the processor:
   ```ts
   avg_heart_rate: activity.average_heartrate ? Math.round(activity.average_heartrate) : null,
   ```
2. Migrate column type to allow floats (already true for `distance`, `duration_minutes`, `calories_burned`):
   ```sql
   ALTER TABLE exercise_entries ALTER COLUMN avg_heart_rate TYPE numeric;
   ```

Optional: only count successfully-inserted activities in `activitiesProcessed`, or expose `activitiesFailed` so the caller can see partial failures.

## Reproduction

1. Connect a Strava account that has activities with `average_heartrate` set (any HR-monitor-equipped run/ride).
2. Trigger `POST /api/integrations/strava/sync`.
3. Observe success response, no rows in `exercise_entries`, errors in server log.

## Workaround applied locally

`ALTER TABLE` + in-place patch of `stravaDataProcessor.ts` with `Math.round`. After both changes, the same 4 activities import correctly with `avg_heart_rate` values `127, 108, 112, 133`.

stellakamikaze:
Fixes #1256.

## Problem

`stravaDataProcessor.ts` passes `activity.average_heartrate` (a float, e.g. `126.9`) directly to `exercise_entries.avg_heart_rate`, which is declared as `integer` in schema. Postgres rejects every such INSERT with:

```
invalid input syntax for type integer: "126.9"
```

So **no Strava activity with heart-rate data ever lands in the diary** for affected users.

The failure is invisible to clients: `activitiesProcessed` in `stravaService.ts` is incremented per loop iteration regardless of insert success, so the API returns `{"success":true,"activitiesProcessed":N}` even when zero rows were written. Locally I confirmed every activity failing this way before the fix.

## Fix

One-line change: round `average_heartrate` to integer before passing it to the persistence layer. Keeps the column type unchanged, matches what the diary actually displays, and there's no semantically meaningful precision lost on bpm averages.

## Verification

After applying the patch and re-running a full backfill (`POST /api/integrations/strava/sync` with `startDate: "2010-01-01"`), all 312 historical activities imported successfully — `avg_heart_rate` correctly rounded (e.g. `127`, `108`, `133`).

## Out of scope (suggested follow-ups)

- Surface partial failures in the sync response (e.g. `activitiesFailed`) so silent insert errors don't go unnoticed.
- Consider migrating `exercise_entries.avg_heart_rate` to `numeric` for consistency with `distance`, `duration_minutes`, `calories_burned` — would also let the field carry more precision if a future provider needs it.

### DIFF
--- a/SparkyFitnessServer/integrations/strava/stravaDataProcessor.ts
+++ b/SparkyFitnessServer/integrations/strava/stravaDataProcessor.ts
@@ -147,8 +147,11 @@ async function processStravaActivities(
         duration_minutes: durationMinutes,
         calories_burned: caloriesAuto,
         distance: distanceKm,
+        // Strava returns average_heartrate as a float (e.g. 126.9), but
+        // exercise_entries.avg_heart_rate is INTEGER → Postgres rejects the insert.
+        // Round before passing through. See issue #1256.
         // @ts-expect-error TS(2339): Property 'average_heartrate' does not exist on typ... Remove this comment to see the full error message
-        avg_heart_rate: activity.average_heartrate || null,
+        avg_heart_rate: activity.average_heartrate ? Math.round(activity.average_heartrate) : null,
         // @ts-expect-error TS(2339): Property 'moving_time' does not exist on type 'nev... Remove this comment to see the full error message
         notes: `Synced from Strava. Type: ${sportType}${activity.moving_time ? `. Moving time: ${Math.round(activity.moving_time / 60)}min` : ''}${activity.total_elevation_gain ? `. Elevation: ${activity.total_elevation_gain}m` : ''}`,
         entry_source: 'Strava',


CodeWithCJ (approved):

CodeWithCJ:
Thanks for your PR. Unfortunately,  I can't merge this PR. I understand this is one liner change so you cleared all the questionaries. 
CI Tests & PR Checks are failing and this is the reason we want everyone submitting PR to answer them so the code merged aligns with existing architecture and standards.

CodeWithCJ:
Just checking when will you be able to fix the CI test errors

CodeWithCJ:
made changes in here https://github.com/CodeWithCJ/SparkyFitness/pull/1273

PR #1306, opened 2026-05-22
fix: short-circuit API-key rate-limit lockout (#1302)

### LINKED ISSUE #1302: [Bug]: Android app gets rate-limited under normal use - default API-key limit (100/60s) is too low, mobile retry policy amplifies it
### What happened?

The Android app intermittently appears "disconnected" from the server: every authenticated request returns HTTP 429 for ~60 seconds, then recovers, then fails again on the next session. The web UI is unaffected because it authenticates via session cookie, not an API key.

The cause is the interaction of two things in this repo:

1. **Server: every authenticated request ticks the API-key rate-limit bucket.** `SparkyFitnessServer/middleware/authMiddleware.ts:54` calls `auth.api.getSession({ headers: req.headers })` on every authenticated request. When the request carries a Bearer API key (the Android app's auth mode), the Better Auth `api-key` plugin treats that call as a verification tick and increments `api_key.request_count`. The plugin is configured in `SparkyFitnessServer/auth.ts:107-121` with defaults `maxRequests: 100`, `timeWindow: 60_000` - i.e. **100 requests per minute, per key**. Any screen that fans out 5–10 react-query calls plus a background sync exhausts that budget in normal use.

2. **Mobile: react-query retries on 4xx, including 429.** `SparkyFitnessMobile/src/hooks/queryClient.ts:3-13` sets the global queries default to `retry: 2` with exponential backoff. Because that's an integer, not a predicate, react-query retries every failure — including 429s, which can't possibly succeed during the cooldown. Each genuine 429 becomes three 429s (initial + ~2s + ~6s), and the retry schedule ignores the server's `Retry-After` header entirely. The codebase already has 429-aware handlers in specific screens (`OnboardingScreen.tsx:308`, `ReauthModal.tsx:190`, `ServerConfigModal.tsx:239`) - the pattern is locally established, just not applied at the queryClient default.

Either fix on its own helps; both together restore normal mobile use without manual DB tuning.

I traced this on my self-hosted instance and manually ran `UPDATE api_key SET rate_limit_max = 600, request_count = 0` on the affected key to keep the app usable.

The Better Auth `api-key` plugin itself is **not at fault** - its rate-limit math (`packages/api-key/src/rate-limit.ts`) is correct; `lastRequest` is not updated on rejected requests, so the window does drain naturally once the client backs off. The problem is the SparkyFitness-side defaults and the mobile retry policy.

### Steps to reproduce

1. Install the Android app and finish onboarding (creates an API key with default `rate_limit_max=100`, `rate_limit_time_window=60000`).
2. Open the app and navigate the dashboard / log a meal / open settings — anything that triggers normal screen fan-out within the same minute.
3. After ~30–60 seconds of normal use, every authenticated request starts returning 429 `{ "error": "Rate limit exceeded." }`.
4. Backend logs show repeated `[ERROR] Error checking Better Auth identity: [APIError: Rate limit exceeded.] { code: 'RATE_LIMITED', details: { tryAgainIn: ~59000 } }`.
5. App appears "disconnected" until the 60-second window lapses without traffic. Reopening the app immediately starts the cycle again.

Inspect the `api_key` row at the time of the failure:

```sql
SELECT name, rate_limit_max, rate_limit_time_window, request_count, last_request
FROM api_key;
-- request_count will be sitting at exactly rate_limit_max (100)
```

### Expected behavior

A normal mobile session should not trip the API-key rate limit. Two complementary fixes:

**Server side** — pick one or both:

- (a) Raise the api-key plugin defaults in `SparkyFitnessServer/auth.ts` so chatty SPA/mobile traffic isn't rejected (e.g. `maxRequests: 600`).
- (b) Cache the session lookup in `SparkyFitnessServer/middleware/authMiddleware.ts` for a short TTL (e.g. 5–10s keyed on a hash of the API key) so a single screen fan-out doesn't burn 10+ verification ticks.

**Mobile side** — replace the integer `retry: 2` with a predicate that excludes 4xx:

```ts
// SparkyFitnessMobile/src/hooks/queryClient.ts
queries: {
  staleTime: Infinity,
  retry: (failureCount, err) => {
    const status = (err as any)?.statusCode ?? (err as any)?.status;
    if (typeof status === 'number' && status >= 400 && status < 500) return false;
    return failureCount < 2;
  },
  retryDelay: (attemptIndex) => Math.min(1000 * 2 ** attemptIndex, 30000),
},
```

(Optional follow-up: honor `Retry-After` on 429 by parsing the header and using it as `retryDelay` when present.)

Happy to open PRs for any of the above.

Docker Compose

### SparkyFitness version

v0.16.6.2

### Which area is affected?

Server / API, Mobile app (Android)

### Relevant logs or screenshots

```shell
Burst showing a single screen fan-out followed by retries, every request rate-limited:


[2026-05-20T00:14:59.882Z] [INFO]  Incoming request: POST /api/user-preferences/bootstrap-timezone
[2026-05-20T00:14:59.899Z] [INFO]  Incoming request: GET  /api/identity/user
[2026-05-20T00:14:59.901Z] [INFO]  Incoming request: GET  /api/user-preferences
[2026-05-20T00:15:00.948Z] [INFO]  Incoming request: GET  /api/user-preferences   ← retry #1
[2026-05-20T00:15:01.015Z] [INFO]  Incoming request: GET  /api/identity/user      ← retry #1
[2026-05-20T00:15:02.987Z] [INFO]  Incoming request: GET  /api/user-preferences   ← retry #2

[2026-05-20T00:14:10.624Z] [ERROR] Error checking Better Auth identity: [APIError: Rate limit exceeded.] {
  status: 'UNAUTHORIZED',
  body: {
    message: 'Rate limit exceeded.',
    code: 'RATE_LIMITED',
    details: { tryAgainIn: 56835 }
  },
  statusCode: 401
}
[2026-05-20T00:14:13.094Z] [ERROR] Error checking Better Auth identity: [APIError: Rate limit exceeded.] { ... tryAgainIn: 54184 }
[2026-05-20T00:14:15.652Z] [ERROR] Error checking Better Auth identity: [APIError: Rate limit exceeded.] { ... tryAgainIn: 51625 }


`api_key` row state at the time of the incident:


       name        | rate_limit_max | rate_limit_time_window | request_count
-------------------+----------------+------------------------+---------------
 Android App       |            100 |                  60000 |           100
```

### Anything else?

May also affect the iOS mobile app (same mobile code path, presumably same behavior) but I don't have an iOS device to confirm.

Workaround for self-hosters until this is addressed:

```sql
-- Per-key (immediate, only affects this one key):
UPDATE api_key
SET rate_limit_max = 600, request_count = 0, last_request = NULL
WHERE name = 'Android App';
```

Or via env (only affects keys created *after* the redeploy — existing rows keep the values they were created with):

```
SPARKY_FITNESS_API_KEY_RATELIMIT_MAX_REQUESTS=600
SPARKY_FITNESS_API_KEY_RATELIMIT_WINDOW_MS=60000
```

Code references:

- `SparkyFitnessServer/auth.ts:107-121` — api-key plugin defaults.
- `SparkyFitnessServer/middleware/authMiddleware.ts:54, 105-115` — `getSession()` per request + 429 response handling.
- `SparkyFitnessMobile/src/hooks/queryClient.ts:3-13` — global retry default.
- `SparkyFitnessMobile/src/screens/OnboardingScreen.tsx:308`, `src/components/ReauthModal.tsx:190`, `src/components/ServerConfigModal.tsx:239` — existing per-call 429 handlers (pattern proof).

jsandai:
> [!TIP]
> For **Frontend** changes, please run `pnpm run validate` to check for any errors.
> PRs that include tests and clear screenshots are highly preferred!
> Note: AI-generated descriptions must be manually edited for conciseness. Do not paste raw AI summaries.

## Description

The Android app gets HTTP 429s under normal use because the Better Auth api-key plugin's per-key rate-limit bucket (default 100/60s) is ticked on every authenticated request, and the mobile retry layer amplifies every rejection.

Server: cache the resolved session for API-key authenticated requests for 5s, keyed on a sha256 of the key. Skips Better Auth's verify on cache hits within the window. Session-cookie auth is unaffected. 
Mobile: introduce `ApiError(message, statusCode, body)` thrown from the central `apiClient`, and a `queryClient` retry predicate that does not retry on 4xx (incl. 429).

Linked Issue: Closes #1302

## How to Test

1. Check out this branch and run `pnpm install`. Start backend + frontend per `CONTRIBUTING.md`; build/run the mobile app.
2. **Server cache, manual smoke**: log in via the Android app, then open the dashboard / log a meal / open settings repeatedly so the app fans out queries. With the same `api_key.rate_limit_max=100`, you should no longer hit 429 within a minute. Inspect `SELECT request_count FROM api_key;` — the counter should now grow far more slowly (roughly one tick per 5s window of activity per key).
3. **Server cache, automated**:
   - `pnpm --filter sparkyfitnessserver test -- apiKeySessionCache.test.ts` (vitest, unit) — covers hit, miss, TTL expiry, lazy eviction, key isolation, clear.
   - `pnpm --filter sparkyfitnessserver test -- authMiddleware.cache.test.ts` (vitest, integration) — covers the middleware actually using the cache: same-key requests collapse to one `auth.api.getSession()` call, different keys are independent, null sessions are not cached, pre-set `x-api-key` headers are covered alongside the Bearer mapping.
4. **Mobile retry predicate, automated**: `pnpm --filter sparkyfitnessmobile test -- --testPathPattern=queryClient` (jest) — covers `ApiError` 4xx → no retry, 5xx → 2 retries, generic Error → 2 retries.
5. **Mobile apiClient, automated**: `pnpm --filter sparkyfitnessmobile test -- --testPathPattern=apiClient` (jest) — covers 4xx and 5xx both throw `ApiError` with `statusCode` and `body` populated.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

- [ ] N/A — bug fix.

- [ ] N/A — no frontend changes in this PR.

- [x] **[MANDATORY for Backend changes] Code Quality**: Typecheck + lint + tests pass. New file (`utils/apiKeySessionCache.ts`) is TypeScript; no new endpoints (middleware-internal helper); covered by unit tests (`tests/apiKeySessionCache.test.ts`).
- [x] **[MANDATORY for Backend changes] Database Security**: No new tables — no schema or RLS changes.

- [ ] N/A — no UI changes in this PR.

- [x] **[MANDATORY for Mobile changes] Tested on device or emulator**: Built `assembleRelease` from this branch, sideloaded on Android, sustained navigation across Dashboard/Diary/Library and meal logging without any "disconnected" / 429 — see "End-to-end validation" note below for the measured cache hit ratio.

## Screenshots

<details>
<summary>Not applicable - no UI changes</summary>
</details>

## Notes for Reviewers

- **End-to-end validation on my self-hosted instance** — built the server image from this branch, deployed it to the `sparkyfitness` stack with the api-key row left at stock defaults (`rate_limit_max=100`, `time_window=60000`), then sideloaded the `assembleRelease` APK and navigated hard (Dashboard, Diary scroll, Library, meal log) until well past the point that previously triggered a disconnect. Measurements after the test:
  - `api_key.request_count`: **21** of 100 cap
  - "Rate limit exceeded" errors in backend logs since deploy: **0**
  - Bearer-API-key incoming requests since deploy: **145**
  - → cache hit ratio ≈ **85%** (145 incoming requests → 21 verifier ticks), no 429s reached the client.
- **Typed error per @apedley's review** — `ApiError extends Error` in `SparkyFitnessMobile/src/services/api/errors.ts`, mirroring the existing `LoginError` pattern in `authService.ts`. Predicate uses `err instanceof ApiError` rather than property mutation.
- **Cache TTL of 5s** is the worst-case staleness if a key is rotated or disabled — the next request after the window will hit Better Auth again. Happy to make this an env knob (`SPARKY_FITNESS_API_KEY_SESSION_CACHE_TTL_MS`) if you'd prefer.
- **Cache key is a sha256 hash** of the API key — defense-in-depth so a memory dump of the cache doesn't reveal raw keys.
- **`Retry-After` is not yet honored** on the mobile side. Left as a follow-up — the immediate 3× amplification problem is solved by not retrying 4xx at all.
- **Local validation on this branch**:
  - server `pnpm run validate` ✅, `pnpm test` ✅ (653/653 — 4 new tests in `authMiddleware.cache.test.ts`, 6 in `apiKeySessionCache.test.ts`)
  - mobile `pnpm run validate` ✅, `pnpm test` for `queryClient` + `apiClient` + `apiKeySessionCache` ✅ (25/25)
  - Two unrelated test files (`LibraryScreen`, `FoodScanScreen`) timed out under full-suite parallelism but pass cleanly on isolated runs — appears to be pre-existing jest flakiness in those screens, not regressions from this PR.
  - device smoke + measured end-to-end cache effect (see "End-to-end validation on my self-hosted instance" above) ✅

### DIFF
--- a/SparkyFitnessMobile/__tests__/hooks/queryClient.test.ts
+++ b/SparkyFitnessMobile/__tests__/hooks/queryClient.test.ts
@@ -1,0 +1,28 @@
+import { queryClient } from '../../src/hooks/queryClient';
+import { ApiError } from '../../src/services/api/errors';
+
+type RetryFn = (failureCount: number, err: unknown) => boolean;
+
+describe('queryClient default retry predicate', () => {
+  const retry = queryClient.getDefaultOptions().queries!.retry as RetryFn;
+
+  test('does not retry on ApiError 4xx (incl. 429)', () => {
+    expect(retry(0, new ApiError('x', 400))).toBe(false);
+    expect(retry(0, new ApiError('x', 401))).toBe(false);
+    expect(retry(0, new ApiError('x', 404))).toBe(false);
+    expect(retry(0, new ApiError('x', 429))).toBe(false);
+    expect(retry(0, new ApiError('x', 499))).toBe(false);
+  });
+
+  test('retries up to twice on ApiError 5xx', () => {
+    expect(retry(0, new ApiError('x', 500))).toBe(true);
+    expect(retry(1, new ApiError('x', 503))).toBe(true);
+    expect(retry(2, new ApiError('x', 500))).toBe(false);
+  });
+
+  test('retries on generic Error (treat as transient)', () => {
+    expect(retry(0, new Error('network'))).toBe(true);
+    expect(retry(1, new Error('timeout'))).toBe(true);
+    expect(retry(2, new Error('network'))).toBe(false);
+  });
+});
--- a/SparkyFitnessMobile/__tests__/services/api/apiClient.test.ts
+++ b/SparkyFitnessMobile/__tests__/services/api/apiClient.test.ts
@@ -1,0 +1,68 @@
+import { apiFetch } from '../../../src/services/api/apiClient';
+import { ApiError } from '../../../src/services/api/errors';
+
+jest.mock('../../../src/services/storage', () => ({
+  getActiveServerConfig: jest.fn(),
+  proxyHeadersToRecord: jest.fn(() => ({})),
+}));
+
+jest.mock('../../../src/services/LogService', () => ({
+  addLog: jest.fn(),
+}));
+
+jest.mock('../../../src/services/api/authService', () => ({
+  getAuthHeaders: jest.fn(() => ({})),
+  notifySessionExpired: jest.fn(),
+}));
+
+import { getActiveServerConfig } from '../../../src/services/storage';
+
+const mockedGetConfig = getActiveServerConfig as jest.MockedFunction<
+  typeof getActiveServerConfig
+>;
+
+describe('apiFetch error handling', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockedGetConfig.mockResolvedValue({
+      id: 'cfg-1',
+      url: 'https://example.invalid',
+      authType: 'apikey',
+      proxyHeaders: [],
+    } as never);
+  });
+
+  test('throws ApiError with statusCode on 4xx', async () => {
+    global.fetch = jest.fn().mockResolvedValueOnce({
+      ok: false,
+      status: 429,
+      text: async () => 'Rate limit exceeded.',
+      headers: new Headers(),
+    }) as never;
+
+    await expect(
+      apiFetch({ endpoint: '/api/x', serviceName: 's', operation: 'get' })
+    ).rejects.toMatchObject({
+      name: 'ApiError',
+      statusCode: 429,
+      body: 'Rate limit exceeded.',
+    });
+  });
+
+  test('throws ApiError with statusCode on 5xx', async () => {
+    global.fetch = jest.fn().mockResolvedValueOnce({
+      ok: false,
+      status: 503,
+      text: async () => 'Service unavailable',
+      headers: new Headers(),
+    }) as never;
+
+    const promise = apiFetch({
+      endpoint: '/api/x',
+      serviceName: 's',
+      operation: 'get',
+    });
+    await expect(promise).rejects.toBeInstanceOf(ApiError);
+    await expect(promise).rejects.toMatchObject({ statusCode: 503 });
+  });
+});
--- a/SparkyFitnessMobile/src/hooks/queryClient.ts
+++ b/SparkyFitnessMobile/src/hooks/queryClient.ts
@@ -1,10 +1,23 @@
 import { QueryClient } from '@tanstack/react-query';
+import { ApiError } from '../services/api/errors';
 
 export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       staleTime: Infinity, // Only refetch when explicitly triggered or polled
-      retry: 2,
+      // Don't retry on 4xx — those are caller-state errors (auth, validation,
+      // 429 rate-limit) that won't change on a retry. Retrying amplified
+      // 429s by 3× and accelerated the api-key rate-limit lockout in #1302.
+      retry: (failureCount, err) => {
+        if (
+          err instanceof ApiError &&
+          err.statusCode >= 400 &&
+          err.statusCode < 500
+        ) {
+          return false;
+        }
+        return failureCount < 2;
+      },
       retryDelay: (attemptIndex) => Math.min(1000 * 2 ** attemptIndex, 30000),
     },
     mutations: {
--- a/SparkyFitnessMobile/src/services/api/apiClient.ts
+++ b/SparkyFitnessMobile/src/services/api/apiClient.ts
@@ -1,6 +1,7 @@
 import { getActiveServerConfig, proxyHeadersToRecord } from '../storage';
 import { addLog } from '../LogService';
 import { getAuthHeaders, notifySessionExpired } from './authService';
+import { ApiError } from './errors';
 
 export const normalizeUrl = (url: string): string => {
   return url.endsWith('/') ? url.slice(0, -1) : url;
@@ -52,7 +53,7 @@ export async function apiFetch<T>(options: ApiFetchOptions): Promise<T> {
       }
       const errorText = await response.text();
       addLog(`[${serviceName}] Failed to ${operation}: ${response.status}`, 'ERROR', [errorText]);
-      throw new Error(`Server error: ${response.status} - ${errorText}`);
+      throw new ApiError(`Server error: ${response.status} - ${errorText}`, response.status, errorText);
     }
 
     if (response.status === 204 || response.headers?.get('content-length') === '0') {
--- a/SparkyFitnessMobile/src/services/api/errors.ts
+++ b/SparkyFitnessMobile/src/services/api/errors.ts
@@ -1,0 +1,10 @@
+export class ApiError extends Error {
+  constructor(
+    message: string,
+    public statusCode: number,
+    public body?: string
+  ) {
+    super(message);
+    this.name = 'ApiError';
+  }
+}
--- a/SparkyFitnessMobile/src/services/api/exerciseApi.ts
+++ b/SparkyFitnessMobile/src/services/api/exerciseApi.ts
@@ -1,4 +1,5 @@
 import { apiFetch, normalizeUrl } from './apiClient';
+import { ApiError } from './errors';
 import { getActiveServerConfig, proxyHeadersToRecord } from '../storage';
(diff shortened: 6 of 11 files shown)


jsandai:
@gemini-code-assist - applied in `1e8855f`. Quick note on framing and the practical reasoning, for the human reviewer's sake:

**Framing nit (minor):** I wouldn't call this a "memory leak" - that implies entries grow without bound for normal traffic. In practice the cache only stores **successful** session lookups (`if (session && session.user && apiKeyToken)` guard in the middleware), so invalid keys/token spray from bots never enter it. Steady-state size is bounded by the number of *valid* API keys, which for a self-hosted SparkyFitness instance is `O(active users)` - typically tens, not thousands. The bigger question Gemini's review is really raising is "is there a defensive ceiling against pathological worst cases" - and the answer should be yes, even if the realistic ceiling is low.

**What I did:**

- Added `MAX_API_KEY_SESSION_CACHE_SIZE = 1000` constant.
- On `setCachedSession`, if the cap is reached and the incoming key isn't already in the map, evict the oldest insertion (FIFO via `cache.keys().next().value`).
- Re-setting an existing key is a no-op for size and does not evict — so a high-traffic instance with N < 1000 active keys never churns.
- 1000 entries × ~2 KB/entry ≈ 2 MB worst-case footprint. Comfortable.

**Why FIFO is fine here (not LRU):** the 5-second TTL means the oldest insertion is almost always already expired by the time we hit the cap. A "true" LRU would require `delete + set` on every `get` to move entries to the back, which adds a write on the hot path with no real benefit at this TTL. FIFO has the same practical effect with a single touch on overflow.

**Tests:** two new cases in `apiKeySessionCache.test.ts`:

- `caps cache size and evicts the oldest entry on overflow (FIFO)` - inserts `MAX + 50` distinct keys, asserts size stays at `MAX` and that the first 50 are evicted.
- `does not evict on re-set of an existing key (no churn for steady traffic)` - refreshing an existing entry doesn't bump size or evict.

Full server suite: 655 / 655 ✅ (+2 from this change), validate ✅.

apedley:
This looks great. The only thing I’d add is including the two fetch calls that use multipart because the server demands it. In exerciseApi.ts and healthDataApi.ts they throw a raw Error. Since it’s just two it’s fine to just have them throw ApiError instead of trying to modify apiFetch for them.

jsandai:
Good catch, addressed in f474f4f. Three throw sites switched to ApiError(message, statusCode, body):

- SparkyFitnessMobile/src/services/api/exerciseApi.ts - create exercise (multipart) + update exercise (multipart)
- SparkyFitnessMobile/src/services/api/healthDataApi.ts - the 4xx branch in the chunked-upload retry wrapper

The 5xx branch in healthDataApi.ts still throws a plain Error since it's caught and converted to a retry inside the same function — leaving it to keep the diff minimal, but happy to make that one consistent too if you'd prefer.

Mobile pnpm run validate ✅, affected test suites (exerciseApi, healthDataApi, apiClient, queryClient) ✅ (130 / 130).

apedley:
Looks good!

CodeWithCJ (approved):

PR #1343, opened 2026-05-27
fix(mcp): publish proper inputSchema for sparky_manage_* tools

ATECHPCS:
## Problem

The three `sparky_manage_*` tools publish empty `inputSchema` to MCP clients in `tools/list`:

```jsonc
"inputSchema": { "type": "object", "properties": {} }
```

Reproduction against any deployment (replace `$URL` / `$TOKEN`):

```bash
curl -sS -X POST $URL -H "Authorization: Bearer $TOKEN" \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}' \
  | jq '.result.tools[] | {name, props: (.inputSchema.properties | length)}'
```

Result: `sparky_manage_food`, `sparky_manage_exercise`, `sparky_manage_checkin` all show `props: 0`. The other tools (e.g. `sparky_inspect_schema`, `sparky_get_health_summary`, `sparky_analyze_food_image`) all show populated properties and a `$schema` field.

## Cause

All three broken tools pass a `z.discriminatedUnion("action", [...])` to `server.registerTool({ inputSchema: ... })`. The MCP TypeScript SDK at `^1.27.1` converts `z.object()` to JSON Schema correctly — you can see `$schema` and populated `properties` on the working tools — but silently falls back to `{type:"object",properties:{}}` for any other Zod type.

Runtime validation still works (Zod errors come back via `isError:true`), but clients can't see fields. Models guess (`name` instead of `food_name`, `protein_g` instead of `protein`, `date` instead of `entry_date`, `snack` instead of `snacks`) and every call fails Zod validation upstream — opaque to most clients.

## Fix

For each of the three tools:

1. Add `export const manage*Input = z.object({...})` in the matching schemas file. The flat shape lists every field any action variant accepts, all optional, each with `.describe()` text indicating which actions need it.
2. In the tool file, import the new shape and pass it as `inputSchema`.
3. Replace the unchecked cast at the top of the handler with `manage*Schema.safeParse(rawArgs)`, returning the structured Zod error via `ERRORS.VALIDATION` if invalid.

The discriminated union still owns runtime correctness; the flat shape only exists to give MCP clients the schema they need.

## Files

| File | Change |
|---|---|
| `SparkyFitnessMCP/src/schemas/food.ts` | +74 — adds `manageFoodInput` |
| `SparkyFitnessMCP/src/schemas/exercise.ts` | +58 — adds `manageExerciseInput` |
| `SparkyFitnessMCP/src/schemas/checkin.ts` | +53 — adds `manageCheckinInput` |
| `SparkyFitnessMCP/src/tools/food.ts` | +10 / -3 — swap `inputSchema`, `safeParse` |
| `SparkyFitnessMCP/src/tools/exercise.ts` | +10 / -3 — same |
| `SparkyFitnessMCP/src/tools/checkin.ts` | +10 / -3 — same |

Total: 6 files, +215 / -9.

## Test plan

- [x] `npm install && npm run build` in `SparkyFitnessMCP/` (the submit script does this; PR will fail to push if it doesn't compile)
- [ ] Start the MCP server locally; hit `tools/list`; confirm all three `manage_*` tools now publish non-empty `inputSchema` with an `action` enum
- [ ] For each tool, call a valid per-action payload (e.g. `{"action":"create_food","food_name":"Test","calories":100,"protein":5,"carbs":10,"fat":3}`) — should succeed exactly as before
- [ ] Call with an invalid payload (missing required field, wrong type, unknown action) — `safeParse` returns the structured Zod error via `ERRORS.VALIDATION` instead of the previous default-fallback message
- [ ] Verify in an MCP client (Claude Desktop / Cline / Cursor / etc.) that the tools now expose their field structure on introspection

### DIFF
--- a/SparkyFitnessMCP/src/schemas/checkin.ts
+++ b/SparkyFitnessMCP/src/schemas/checkin.ts
@@ -97,3 +97,56 @@ export const manageCheckinSchema = z.discriminatedUnion("action", [
 ]);
 
 export type ManageCheckinInput = z.infer<typeof manageCheckinSchema>;
+
+
+// Flat input shape published to MCP clients as `inputSchema`. See comment on
+// manageFoodInput in ./food.js for the rationale. Runtime validation still
+// uses manageCheckinSchema in the tool handler via safeParse.
+export const manageCheckinInput = z.object({
+  action: z.enum([
+    "log_biometrics",
+    "log_custom_metric",
+    "list_categories",
+    "create_category",
+    "log_mood",
+    "log_fasting",
+    "log_sleep",
+    "list_checkin_diary",
+    "get_fasting_status",
+    "get_biometrics_history",
+  ]).describe("Action to perform. Each value selects a per-action signature; see the tool description for required fields per action."),
+  entry_date: dateSchema.optional().describe("Date for the entry (YYYY-MM-DD)"),
+  // biometrics
+  weight: z.coerce.number().min(0).optional().describe("Weight value"),
+  weight_unit: weightUnitEnum.optional().describe("Unit for weight (defaults to kg)"),
+  steps: z.coerce.number().int().min(0).optional().describe("Daily step count"),
+  height: z.coerce.number().min(0).optional().describe("Height value"),
+  height_unit: heightUnitEnum.optional().describe("Unit for height"),
+  neck: z.coerce.number().min(0).optional().describe("Neck measurement"),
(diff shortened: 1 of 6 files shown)


CodeWithCJ (approved):

CodeWithCJ (approved):

CodeWithCJ:
Thanks for the PR. For future PRs, please don't clear PR questionaries as we have setup some PR validation that would fail otherwise.

PR #1381, opened 2026-05-30
feat(mobile): add Camera shutter toggle in App Settings

### LINKED ISSUE #1372: [Feature]: Add "Sounds" toggle to App Settings

The mobile photo capture flow plays the system shutter sound on every shot. In a restaurant, an office or a quiet kitchen, that sound is loud and intrusive. There's currently no in-app way to silence it short of putting the device in full silent mode, which also kills notifications, alarms, and other audio.

Add a **"Sounds"** toggle to `AppSettingsScreen`, defaulted **on** to match current behavior. When toggled off, `CameraView` in `FoodScanScreen` reads the preference and passes `mute={true}` to `expo-camera`, which silences the shutter without affecting any other system audio.

This mirrors the existing **Haptics** toggle in the same screen one-to-one: same row layout, same default-on, same on-means-it-plays semantics, same `AsyncStorage`-backed preference layer. The shutter toggle introduces a `useSoundsEnabled` hook and a `setSoundsEnabled` setter in a sibling `src/services/sounds.ts` (or directly in `haptics.ts`'s neighborhood) that mirror `useHapticsEnabled` / `setHapticsEnabled` shape-for-shape.

Mobile app (iOS), Mobile app (Android)

Nice to have

- **System silent / Do Not Disturb**: kills notifications, alarms, and other audio too. Heavy hammer.
- **Phone volume slider**: shutter sound is enforced at system level on some Android devices regardless of media volume.
- **A "Mute camera shutter" toggle**: narrower naming, but breaks symmetry with the existing Haptics toggle and forces the user to parse a negation (on = silence). The generic "Sounds" toggle matches OS-level naming conventions on iOS and Android and pattern-matches the Haptics row.

### Additional context

Implementation is small: one new boolean preference (mirroring `hapticsEnabled` plumbing), one read in `FoodScanScreen`, and the `mute` prop on `<CameraView />`. Tests can mock the preference and assert the prop value.

Happy to open the PR once this is reviewed.

jsandai:
## Description

The Android camera shutter fires a loud system click when capturing food photos. Resolves #1372 by adding an opt-out under Settings → App Settings.

New `src/services/sounds.ts` (storage key `@HealthConnect:soundsEnabled`, default on) following the same shape as `src/services/haptics.ts`. A "Camera shutter" Switch row is added to `AppSettingsScreen` directly under Haptic Feedback. `FoodScanScreen` reads the preference via `useSoundsEnabled()` and passes `shutterSound: soundsEnabled` into `expo-camera`'s `takePictureAsync` at both capture call sites (label scan + food photo).

`shutterSound` is the documented `expo-camera` knob for the capture sound and works on both iOS and Android (on iOS it invokes Apple's private `AudioServicesDisposeSystemSoundID(1108)` path inside `expo-camera`; on Android it disables the `MediaActionSound` the OS plays on capture).

Linked Issue: Closes #1372

## How to Test

1. Check out this branch and run `pnpm install && pnpm start` from `SparkyFitnessMobile/`.
2. Open the app, go to Settings → App Settings. You should see a new "Camera shutter" row directly under Haptic Feedback. Default is on.
3. Open Add → Scan, then take a photo in either Label or Photo mode. With the toggle on you hear the normal shutter sound. With the toggle off the shutter is silent.
4. Force-quit the app and reopen. The toggle state persists.

## PR Type

- [x] New Feature
- [ ] Issue (bug fix)
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

- [x] **[MANDATORY for new feature] Alignment**: I have raised a GitHub issue and it was reviewed/approved by maintainers or it was approved on Discord. (Issue #1372, acknowledged by @apedley.)

- [x] **[MANDATORY for Mobile changes] Tested on device or emulator**: I have verified the changes work on iOS or Android. - Verified on Samsung Galaxy S26 Ultra: shutter audible when toggle is on, silent when toggle is off, state persists across force-quit. Changes should also work on iOS – `expo-camera`'s `shutterSound: false` triggers `AudioServicesDisposeSystemSoundID(1108)` in `CameraPhotoCapture.swift` – but I do not have iOS hardware to verify directly. Note that iOS shutter mute is region-restricted by Apple in Japan and South Korea regardless of the flag.

## Screenshots

<details>

### Before

Screenshot_20260530_122753_SparkyFitness

### After

Screenshot_20260530_122846_SparkyFitness

</details>

### Why camera-only and not a full app-sounds toggle

The issue thread (#1372) framed this as both silencing the camera shutter *and* the rest-timer notification. I initially attempted a broader "App sounds" toggle that would also cover rest-timer notification audio, but ran into a foreground-delivery limitation I could not reliably solve from the JS layer.

What I found, on Samsung One UI 7 (S26 Ultra) and an older S6 tablet, both on the public v0.16.6.3 release with no modifications:

- The workout rest-timer notification fires correctly when the app is **backgrounded** (sound + haptic).
- When the app is in the **foreground**, only the haptic fires. The configured channel sound never plays.

This reproduces against the unmodified release and is independent of any changes in this PR – it appears to be an Android foreground-delivery behaviour rather than something a per-call `sound: true` flag on the scheduled notification can override from JS. Properly fixing it likely means either an in-process audio playback path (e.g. `expo-av`) layered onto the existing notification flow, or letting the OS handle it via the notification channel and accepting current foreground behaviour.

Rather than ship a sounds toggle that visually claims to silence the rest timer while the rest timer was *already* not audible in the foreground, I narrowed scope to the one place a user can clearly hear a sound today: the camera shutter. Labeling the toggle "Camera shutter" keeps the UI honest about what it does, and leaves the foreground notification audio question for a future, focused PR.

If you would prefer a different name (e.g. "Mute camera shutter") or want this nested under a future "Sounds" group, happy to adjust.

### DIFF
--- a/SparkyFitnessMobile/App.tsx
+++ b/SparkyFitnessMobile/App.tsx
@@ -88,6 +88,7 @@ import {
 } from './src/services/autoSyncCoordinator';
 import { initializeTheme } from './src/services/themeService';
 import { initializeHaptics } from './src/services/haptics';
+import { initializeSounds } from './src/services/sounds';
 import { loadActiveDraft, clearDraft } from './src/services/workoutDraftService';
 import { addLog, initLogService } from './src/services/LogService';
 import { initNotifications } from './src/services/notifications';
@@ -430,6 +431,7 @@ function AppContent() {
     // Initialize theme from storage on app start
     initializeTheme();
     initializeHaptics();
+    initializeSounds();
 
     // Reset the auto-open flag on every app start
     const initializeApp = async () => {
--- a/SparkyFitnessMobile/__tests__/services/sounds.test.ts
+++ b/SparkyFitnessMobile/__tests__/services/sounds.test.ts
@@ -61,4 +61,18 @@ describe('sounds service', () => {
 
     expect(getSoundsEnabled()).toBe(false);
   });
+
+  it('does not let initializeSounds overwrite a user toggle that lands mid-flight', async () => {
+    // Storage has "true". Start init but don't await it yet — it is now sitting
+    // on `await AsyncStorage.getItem(...)`.
+    await AsyncStorage.setItem(STORAGE_KEY, 'true');
+    const initPromise = initializeSounds();
+
+    // While the storage read is in flight, the user toggles off.
+    await setSoundsEnabled(false);
+
+    // Now let initializeSounds resolve. It must not clobber the user value.
+    await initPromise;
+    expect(getSoundsEnabled()).toBe(false);
+  });
 });
--- a/SparkyFitnessMobile/__tests__/services/sounds.test.ts
+++ b/SparkyFitnessMobile/__tests__/services/sounds.test.ts
@@ -1,8 +1,11 @@
 import AsyncStorage from '@react-native-async-storage/async-storage';
+import { renderHook, act } from '@testing-library/react-native';
 import {
   initializeSounds,
   setSoundsEnabled,
   getSoundsEnabled,
+  useSoundsEnabled,
+  __resetSoundsStateForTests,
 } from '../../src/services/sounds';
 
 const STORAGE_KEY = '@HealthConnect:soundsEnabled';
@@ -10,9 +13,7 @@ const STORAGE_KEY = '@HealthConnect:soundsEnabled';
 describe('sounds service', () => {
   beforeEach(async () => {
     await AsyncStorage.clear();
-    // Reset module-level state to default (enabled) before each test.
-    await setSoundsEnabled(true);
-    await AsyncStorage.clear();
+    __resetSoundsStateForTests();
   });
 
   it('defaults to enabled when nothing is persisted', async () => {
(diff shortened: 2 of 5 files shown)


jsandai:
@gemini-code-assist - All three findings addressed in b9e6a1f:

- HIGH (mid-flight race in `initializeSounds`): added the `if (!initialized && saved !== null)` guard before applying the stored value, and the matching `if (!initialized)` guard in the `finally` block. A user toggle that lands while the AsyncStorage read is in flight now wins, where previously the late storage read could still clobber it.
- MEDIUM × 2 (`useSyncExternalStore`): replaced the manual `useState`/`useEffect` subscription in `useSoundsEnabled` with `useSyncExternalStore`. Imports updated. This is the React 18+ idiom for external stores and handles tearing under concurrent rendering.
- Added a new test (`does not let initializeSounds overwrite a user toggle that lands mid-flight`) that reproduces the race the HIGH finding flagged: starts `initializeSounds()` without awaiting, calls `setSoundsEnabled(false)` while the storage read is pending, then awaits the init promise and asserts the user value survived.

Full suite: 143 suites / 2271 tests passing. Typecheck + lint clean.

apedley:
Looks good. Verified this works on ios.

CodeWithCJ (approved):

Issue #1510, opened 2026-06-11
[Bug]: Auto-Scale is force-enabled on all foods (and the base variant), not just online imports


jsandai:
### What happened?

The "Auto-scale Online Database Imports" preference is documented as: "When enabled, the Auto-Scale checkbox will be on by default when editing foods imported from any online database, so nutrition values scale automatically when you change the serving size." So by its own description it should affect only foods imported from an online database.

In practice, with the preference enabled, the food form force-enables Auto-Scale on every variant of every food, including plain custom foods that were never imported, and including the base/default variant where auto-scale has nothing to scale from. While Auto-Scale is ticked, every nutrition input on that variant is disabled, so the macros are greyed out and cannot be edited until Auto-Scale is manually unticked. The variant's own stored Auto-Scale state is ignored.

### Steps to reproduce

1. Enable the "Auto-scale Online Database Imports" preference in Settings, Preferences.
2. Open a plain custom food (one that was never imported from an online database) in the food form, via Foods then Edit food.
3. Observe that the variant shows Auto-Scale ticked and all of its calorie and macro fields are greyed out and not editable, even though this food was not imported.
4. Untick Auto-Scale on the variant. The fields become editable again.

### Expected behavior

Per the preference description, Auto-Scale should default on only for foods imported from an online database, not for plain custom foods. In addition, the base/default variant's nutrition fields should always be editable, since auto-scale has nothing to scale from on the base variant. The variant's own stored Auto-Scale state should be respected rather than overridden on load.

### How are you running SparkyFitness?

Docker Compose

### SparkyFitness version

v0.16.9

### Which area is affected?

Web frontend

### Relevant logs or screenshots

```shell
The macro fields render greyed out with Auto-Scale ticked.
```

### Anything else?

Root cause (traced in the frontend):

- `SparkyFitnessFrontend/src/components/FoodSearch/NutrientFormGrid.tsx` derives `const isLocked = variant.is_locked ?? false;` and applies `disabled={isLocked}` to every nutrition input, so a variant with `is_locked` true has all of its macro fields disabled.
[...]

PR #1511, opened 2026-06-11
feat: Add Duplicate action to the food Actions menu

### LINKED ISSUE #1488: [Feature]: Duplicate a food from the Actions menu

Many foods come in variants or flavours where the nutrition differs only slightly (for example a protein powder, bar, or sauce sold in several flavours). Adding each one means re-entering the full set of macros from scratch, even though most of the values are identical across flavours. There is currently no way to start a new food from an existing one.

Add a "Duplicate" action to a food's three-dot Actions menu (alongside View details, Edit food, Share with public, Delete food). Selecting it opens the food creation form pre-filled from the source food: the name suffixed with " (copy)", the brand, and all unit variants with their full nutrition. The user adjusts whatever differs for the new flavour and saves it as a new food. The original food is not modified, and the copy is private by default.

Web frontend, Nutrition tracking

- Re-entering the food from scratch each time (current behavior), which is slow and error-prone for variant-heavy foods.
- Editing the original food, which is not viable because saving would overwrite the original rather than create a new one.

### Additional context

This can reuse the existing create-food flow, so no server change is needed: Duplicate opens the same form used for adding a custom food, pre-filled, and saves through the existing create path. Food and variant IDs would be stripped so the result is a genuine new food rather than an edit of the original, and the public/shared flag would reset to private.

Happy to open a PR once this issue is reviewed.

jsandai:
## Description

Foods that come in flavour or size variants currently have to be re-entered from scratch even when only one or two macros differ. This PR adds a Duplicate action so an existing food can be cloned and adjusted.

A new "Duplicate food" item in the food Actions menu opens the existing create-food form pre-filled from the source food (name suffixed with "(copy)", brand, and all unit variants with their nutrition). Clearing the food id routes the save through the existing food-creation path, which assigns fresh food and variant ids, so the original is never modified, the copy is owned by the current user, and it defaults to private. Source identity is dropped from the copy (food and variant ids, barcode, and provider linkage) so it saves as a clean manual food. Clearing the barcode in particular matters: createFood deduplicates by barcode and would otherwise return the original food instead of creating a new one. Variants are pre-fetched and attached inline because an id-less food does not trigger the server-side variant load. No server change.

Linked Issue: Closes #1488

## How to Test

1. Check out this branch and run `pnpm install` at the repo root.
2. Start the stack (for example `docker compose -f docker/docker-compose.dev.yml up`, or run the frontend with `pnpm --filter sparkyfitnessfrontend run dev` against a running backend).
3. Go to Foods and open the three-dot Actions menu on any food, then choose "Duplicate food".
4. Confirm the form opens pre-filled with the name plus " (copy)", the brand, and every unit variant with its nutrition.
5. Adjust a value and save. Confirm a new private food is created and the original food is unchanged.
6. Duplicate a food that has a barcode (for example an imported product) and confirm a genuine new food is created rather than the original being returned.
7. Duplicate a System or Family food you do not own and confirm the copy saves as your own private food.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

### Before

2026-06-11 17_30_31-Greenshot

### After

2026-06-11 16_04_36-Greenshot

</details>

## Notes for Reviewers

The Duplicate action reuses the existing create-food flow with no server change. The copy is seeded with no source identity (ids, barcode, provider fields stripped) so it is a fresh manual food owned by whoever duplicates it. Stripping the barcode is deliberate: createFood matches an incoming barcode to an existing food and returns it, so without this a duplicate of a barcoded food would be a silent no-op.

Duplicate is intentionally available for all foods, including System and Family foods the user does not own, since the result is always a new private food owned by the current user. That is the most useful case (cloning a public food to customise it). Happy to gate it behind ownership if preferred.

While testing this, I hit a separate, pre-existing issue: with the "Auto-scale Online Database Imports" preference enabled, the create/edit form locks the macro fields on every variant (including the base variant) so they cannot be edited until Auto-Scale is unticked. It is unrelated to this change and is reported separately in #1510. Calling it out so it does not get attributed to this PR.

Validation: `pnpm run validate` passes, and the existing food form test suite passes.


jsandai:
@gemini-code-assist

Applied. The duplicate now falls back to the default variant if the fetched variants array comes back empty, so the create form is never seeded without a variant.

In practice the fetched array always contains at least the default variant, so this only affects the empty-fetch edge case; the common path is unchanged.

CodeWithCJ (approved):

PR #1513, opened 2026-06-12
Keep nutrition fields editable when Auto-Scale is on

### LINKED ISSUE #1510: [Bug]: Auto-Scale is force-enabled on all foods (and the base variant), not just online imports

jsandai:
## Description

With the auto-scale preference on, opening any food to edit showed every nutrition field greyed out and uneditable until Auto-Scale was unticked on each variant. This made it look like the values could not be changed.

Decouple the Auto-Scale checkbox from the field lock. Auto-Scale still rescales a variant's nutrition values when you change its serving size, but the inputs are no longer disabled, so you can auto-scale to roughly the serving you want and then type over any value. The form already re-captures the scaling base on a manual edit, so a later serving-size change scales from your edited value rather than the original. The variant-scaling logic in useFoodForm is unchanged, including the guard that does not scale variants from an AI unit conversion, so this is a UI-only change in NutrientFormGrid plus a wording update.

Linked Issue: Closes #1510

## How to Test

1. Check out this branch and run `pnpm install` at the repo root, then start the stack.
2. In Settings, Preferences, enable Auto-Scale (or open a food whose variant has Auto-Scale checked).
3. Open a food to edit. Confirm the calorie and macro fields are editable while Auto-Scale is checked.
4. Change the serving size and confirm the values rescale.
5. Type a new value into a macro field, then change the serving size again, and confirm it scales from your edited value.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

## Screenshots

<details>

### Before

2026-06-11 22_28_03-Greenshot

2026-06-11 22_37_13-Greenshot

### After

2026-06-11 22_28_12-Greenshot

2026-06-11 22_35_49-Greenshot

</details>

## Notes for Reviewers

This follows the discussion on #1510. Auto-Scale no longer disables the inputs, so the on/off default stops being a blocker either way. I left the preference default unchanged (on); happy to flip it to off as well if preferred, it is a one-line change.

The scaling behaviour itself is untouched: useFoodForm already sets an edited value as the new scaling base and already skips scaling for AI-unit-converted variants. Only the field disable in NutrientFormGrid and the preference wording changed. Added a hook test covering the edit-then-rescale case.

### DIFF
--- a/SparkyFitnessFrontend/src/components/FoodSearch/NutrientFormGrid.tsx
+++ b/SparkyFitnessFrontend/src/components/FoodSearch/NutrientFormGrid.tsx
@@ -50,7 +50,6 @@ function NutrientInput({
   value,
   step = '0.1',
   decimals,
-  disabled,
   onChange,
 }: {
   id: string;
@@ -58,7 +57,6 @@ function NutrientInput({
   value: number | undefined;
   step?: string;
   decimals?: number;
-  disabled: boolean;
   onChange: (val: number | undefined) => void;
 }) {
   return (
@@ -71,7 +69,6 @@ function NutrientInput({
         decimals={decimals}
         onValueChange={(value) => onChange(value)}
         placeholder="0"
-        disabled={disabled}
       />
     </div>
   );
@@ -87,7 +84,10 @@ export function NutrientGrid({
   onUpdate,
 }: NutrientGridProps) {
   const { t } = useTranslation();
-  const isLocked = variant.is_locked ?? false;
+  // Auto-Scale (is_locked) controls whether changing the serving size rescales
+  // the nutrition values; it no longer disables the inputs, so the user can
+  // auto-scale and still edit any value directly. A manual edit re-captures the
+  // scaling base in useFoodForm, so a later serving-size change scales from it.
   const update = (field: string) => (val: number | undefined) =>
     onUpdate(variantIndex, field, val);
 
@@ -107,10 +107,7 @@ export function NutrientGrid({
                   onUpdate(variantIndex, 'glycemic_index', val)
                 }
               >
-                <SelectTrigger
-                  id={gridId(variantIndex, 'glycemic_index')}
-                  disabled={isLocked}
-                >
+                <SelectTrigger id={gridId(variantIndex, 'glycemic_index')}>
                   <SelectValue placeholder="Select GI" />
                 </SelectTrigger>
                 <SelectContent>
@@ -141,7 +138,6 @@ export function NutrientGrid({
               }
               step="1"
               decimals={0}
-              disabled={isLocked}
               onChange={update('calories')}
             />
           );
@@ -158,7 +154,6 @@ export function NutrientGrid({
               value={variant[key as NumericFoodVariantKeys]}
               step={cfg.decimals === 0 ? '1' : '0.1'}
               decimals={cfg.decimals}
-              disabled={isLocked}
               onChange={update(key)}
             />
           );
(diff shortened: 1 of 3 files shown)


jsandai:
@gemini-code-assist

Applied the col-span-full suggestion, that is the more correct idiom for spanning the full row across breakpoints.

On the translation files: these two keys are not present in en.json (nor are the sibling autoScaleOpenFoodFacts strings), so i18next falls back to the default text in the t() call and the new strings render as written. I verified this on a build before pushing. Adding the keys to en.json would also be inconsistent with the neighbouring auto-scale setting, which uses the same inline-default pattern, so I have left the translation files unchanged.

CodeWithCJ (approved):

PR #1574, opened 2026-06-19
feat(mobile): copy a logged meal to another day

### LINKED ISSUE #1361: [Feature]: Copy meals on mobile app

Currently you can only copy meals on the web, but not on the mobile app. This means if a user of the app wants to copy meals they have to then go onto the web do it. 

Ability to copy meals on the mobile app.

_No response_

Nice to have

_No response_

### Additional context

_No response_

jsandai:
## Description

The web diary can copy a full meal to another date, but the mobile app cannot, so mobile users have to switch to the web app to do it (issue #1361).

The backend copy endpoint (`POST /food-entries/copy`) already exists, so this is mobile-client only: a copy action on the meal-type detail screen opens a bottom sheet to choose a target date and meal, then calls that endpoint through a new react-query mutation that invalidates the target day. The sheet reuses existing primitives (the diary's date picker and the app's bottom-sheet pattern), and the copy button only appears when the meal has entries and stays disabled until the target differs from the source.

Linked Issue: Closes #1361

## How to Test

1. Log some foods into a meal (for example Dinner) on any day.
2. Open that meal from the diary to reach the Meal Type Detail screen. A copy icon appears at the top-right of the header.
3. Tap it, pick a different target date and/or meal in the sheet, and tap Copy.
4. Verify a "Meal copied" toast appears, the entries now show on the target day and meal, and the source meal is unchanged.
5. Confirm the Copy button is disabled while the target date and meal still match the source.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

- [x] **[MANDATORY for new feature] Alignment**: Raised as issue #1361 and approved on Discord.

- Not applicable: no frontend changes.

- Not applicable: no backend changes.

- [x] **[MANDATORY for UI changes] Screenshots**: Before/After screenshots attached below.

- [x] **[MANDATORY for Mobile changes] Tested on device or emulator**: Verified on a physical Android device (Samsung Galaxy S26 Ultra). Copied both a set of standalone food entries and a built/logged meal from one date to another; the target day populated correctly, the source was unchanged, and the toast fired.

## Screenshots

<details>

### Before

Screenshot_20260616_140418_SparkyFitness

### After

Screenshot_20260616_140156_SparkyFitness

Screenshot_20260616_140205_SparkyFitness

Screenshot_20260616_140214_SparkyFitness

</details>

## Notes for Reviewers

Mobile-only change; no server or schema changes. Meal types are matched by name (case-insensitive) by the existing endpoint, which maps cleanly to the mobile meal-type keys, so no id translation is needed. Scope is intentionally limited to the per-meal copy-to-another-day flow (the core ask in #1361). The web app also has copy-from-yesterday and copy-entire-day; those can be follow-ups.

Verification: 39 unit tests pass (4 new API cases in `foodEntriesApi.test.ts`, 3 new hook cases in `useCopyFoodEntries.test.ts`); `tsc --noEmit` adds no new errors from the changed files; eslint is clean on all changed files.

Interaction with the recent mobile meal-ingredient editing feature (EditLoggedMealScreen): the two are independent. When a copied slot contains a logged meal (a `food_entry_meal` container), the server duplicates the container with a fresh id and links the copied entries to it, so the copy is independently editable and never shares state with the original.

### DIFF
--- a/SparkyFitnessMobile/src/components/CopyMealSheet.tsx
+++ b/SparkyFitnessMobile/src/components/CopyMealSheet.tsx
@@ -92,14 +92,20 @@ const CopyMealSheet = forwardRef<CopyMealSheetRef, CopyMealSheetProps>(
 
     const handleDateChange = useCallback(({ date }: { date: DateType }) => {
       if (!date) return;
-      // The picker hands back a dayjs object; convert via its own toDate() rather
-      // than relying on Date coercion.
-      const jsDate =
-        date instanceof Date
-          ? date
-          : typeof date === 'object' && 'toDate' in date
-            ? date.toDate()
-            : new Date(date);
+      // The picker hands back a dayjs object; convert via its own toDate(). A
+      // 'YYYY-MM-DD' string is parsed with local components so it does not shift
+      // a day in timezones behind UTC (new Date(str) would parse as UTC midnight).
+      let jsDate: Date;
+      if (date instanceof Date) {
+        jsDate = date;
+      } else if (typeof date === 'object' && 'toDate' in date) {
+        jsDate = date.toDate();
+      } else if (typeof date === 'string') {
+        const [year, month, day] = date.split('-').map(Number);
+        jsDate = new Date(year, month - 1, day);
+      } else {
+        jsDate = new Date(date);
+      }
       setTargetDate(toLocalDateString(jsDate));
     }, []);
 
--- a/SparkyFitnessMobile/src/components/CopyMealSheet.tsx
+++ b/SparkyFitnessMobile/src/components/CopyMealSheet.tsx
@@ -20,7 +20,8 @@ import Button from './ui/Button';
 import Icon from './Icon';
 import { useMealTypes } from '../hooks/useMealTypes';
 import { getMealTypeLabel } from '../constants/meals';
-import { formatDateLabel, toLocalDateString } from '../utils/dateUtils';
+import { formatDateLabel } from '../utils/dateUtils';
+import { dayToPickerDate, localDateToDay } from '@workspace/shared';
 import type { CopyFoodEntriesPayload } from '../services/api/foodEntriesApi';
 
 // Render the sheet inside an iOS UIWindow so it sits above any native modal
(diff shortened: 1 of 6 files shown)


CodeWithCJ:
@jsandai  please attache screenshot and also have a look at the gemini comment .tag @apedley  for his review once you are done.

jsandai:
@gemini-code-assist

Good catch on the latent timezone shift. In practice the date picker passes a dayjs object, which is handled by the toDate() branch, so the string path is not currently reached. I have applied the same local-component parsing to the string fallback anyway, so it stays consistent with how the target date is parsed elsewhere in the component and avoids the UTC-midnight day shift if a string is ever passed through.

CodeWithCJ:
@jsandai there is time util in shared package . use that to avoid time shift issues if you are not using that already.

jsandai:
@CodeWithCJ thanks. Switched the date parsing in CopyMealSheet to use dayToPickerDate and localDateToDay from @workspace/shared, so it is consistent with the rest of the codebase and avoids any timezone shift. Screenshots added above. @apedley this is ready for your review when you get a chance.

SparkyFitnessMobile/src/components/CopyMealSheet.tsx
```diff
@@ -0,0 +1,231 @@
...
+    return (
+      <BottomSheetModal
+        ref={bottomSheetRef}
+        enableDynamicSizing
+        backdropComponent={renderBackdrop}
+        containerComponent={sheetContainer}
+        backgroundStyle={{ backgroundColor: surfaceBg }}
+        handleIndicatorStyle={{ backgroundColor: textMuted }}
+      >
+        <BottomSheetScrollView contentContainerClassName="px-5 pb-safe-or-8">
```
apedley: Move the px-5 here to an inner wrapper `View` for an ios padding bug

apedley (changes requested):
Good work as usual. One problem: Currently all non standard meal types are collapsed under "Other" on mobile. Tapping other, copy, and saving results in "Meal copied" but nothing actually saves. I also added a comment about a padding fix for ios

jsandai:
Thanks, both fixed in f10d534.

iOS padding: moved the sheet's horizontal padding off the BottomSheetScrollView content container onto an inner wrapper view, per your comment.

The "Other" copy bug: root cause is that mobile collapses every non-standard meal type into a single synthetic "Other" bucket (getFoodEntryMealTypeKey), so "Other" isn't a real meal type and can mix several custom types together. The copy request matched nothing server-side but still returned success, hence the misleading toast. I've hidden the copy action on the "Other" bucket so it's only offered for concrete meal types. The four standard types (breakfast, lunch, dinner and snack) are unaffected.

On enabling copy for custom meal types (e.g. "Supper", "Pre-workout", etc.): worth noting the target side already supports them, since the "Target meal" chips list every visible meal type. The gap is purely the source: the diary doesn't give custom types their own sections the way the web does, so there's no per-type screen to start a copy from. Making that work means having the mobile diary render a section per visible meal type instead of bucketing them into "Other", which touches the diary's core grouping (MealTypeKey, groupFoodEntriesByMealType, FoodSummary, and the detail-screen routing). 

I deliberately kept that out of this PR rather than fold it in, for two reasons: it changes the diary layout for all mobile users (not just those with custom types), so it's a product/UX decision rather than a copy bug fix; and bundling a diary-wide refactor into a focused feature PR makes it harder to review and slower to land. My preference is to ship copy now with "Other" gated off, and do the per-type sections as a follow-up, at which point copy works for custom types automatically. Happy to open that follow-up, or pull it into this PR instead if you'd rather have it all together.

apedley:
That makes sense. iOS padding is fixed.

@CodeWithCJ this is fine

CodeWithCJ (approved):

Issue #1496, opened 2026-06-10
[Feature]: Unified search


mklatsky:
### What problem does this solve?

It is a little annoying that, when adding a food, I need to choose whether to search from my own database, meals, online or scan.

### Describe the feature you'd like

I would like to see a unified search such that I don't need to select WHERE the data comes from. If I add food in my diary, I simply search for the food and select it to add. 

Updating units, etc should be something done at a different time.

There should be as little friction as possible when adding foods to my diary.

### Which area does this relate to?

Nutrition tracking

### How important is this to you?

Would significantly improve my workflow

### Alternatives you've considered

I currently have a paid subscription top Cronometer - I am hoping to see SparkyFitness have the ease of use that Cronometer has. if it gets 80% there, I'd rather take my subscription $$$ and sponsor or donate to this project.

apedley:
Yeah I’m down for adding this but it has to be done right. I’m thinking possibly searching local database while typing and then searching all external providers on submit/button press. And depending on the ui, ranking becomes a decision. A segmented list may help on mobile.

CodeWithCJ:
@apedley  @jsandai  anything else still pending in this?

apedley:
Nope this is completed

jsandai:
> @apedley  @jsandai  anything else still pending in this? 

Unified search is complete as far as I'm aware. Happy to await Andrew's sign off

ETA: he beat me to it!

PR #1515, opened 2026-06-12
Feature/Use native tabs on iOS

thorethy1:
> [!TIP]
> **Help us review and merge your PR faster!**
> Please ensure you have completed the **Checklist** below.
> For **Frontend** changes, please run `pnpm run validate` to check for any errors.
> PRs that include tests and clear screenshots are highly preferred!
> Note: AI-generated descriptions must be manually edited for conciseness. Do not paste raw AI summaries.

## Description

**What problem does this PR solve?**
It uses expo native tabs to create a liquid glas tab bar.

**How did you implement the solution?**

Divided the tab-bar according to platform:

- iOS now uses @bottom-tabs/react-navigation with

NativeTab.Navigator.

- The existing add tab remains the same route, gets

On iOS 26+ but role: 'search', which makes it

Separate button appears in the new native tabs design.

- The icon remains the existing Add/Plus icon, no

Search UI opens.

- preventsDefault: true and the existing

tabPress.preventDefault() prevent the empty

Add-Tab remains active.

- The AddSheet now recognizes whether it is without selection

Closed, and then navigate back to the

Last real tab.

- Android remains unchanged with the previous

Tab.Navigator with CustomTabBar.

Linked Issue: Closes #1514 

## How to Test

Build a iOS build and test the tab bar.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

- [x] **[MANDATORY - ALL] Integrity & License**: I certify this is my own work, free of malicious code, and I agree to the [License terms](../LICENSE).

**New features only:**

- [x] **[MANDATORY for new feature] Alignment**: I have raised a GitHub issue and it was reviewed/approved by maintainers or it was approved on Discord.

**Frontend changes (`SparkyFitnessFrontend/`):**

- [x] **[MANDATORY for Frontend changes] Quality**: I have run `pnpm run validate` and it passes.
- [ ] **[MANDATORY for Frontend changes] Translations**: I have only updated the English (`en`) translation file.

**Backend changes (`SparkyFitnessServer/`):**

- [x] **[MANDATORY for Backend changes] Code Quality**: I have run typecheck, lint, and tests. New files use TypeScript, new endpoints have Zod schemas, and new endpoints include tests.
- [ ] **[MANDATORY for Backend changes] Database Security**: I have updated `rls_policies.sql` for any new user-specific tables.

**UI changes (components, screens, pages):**

[...]

Issue #1079, opened 2026-04-05
OpenFoodFacts search broken: migrate to Search-a-licious API


dtgagnon:
## Problem

The OpenFoodFacts legacy search endpoint (`world.openfoodfacts.org/cgi/search.pl`) is returning HTTP 503 errors globally. This affects all text-based food searches using the OpenFoodFacts provider. Barcode lookups (`/api/v2/product/{barcode}`) still work fine.

It's not a rate limiting or a configuration issue. 503 error occurs on the very first request, from any IP, with any User-Agent, including browser UAs. The legacy search infrastructure appears to be down.

## Potential root cause

Per the [official OFF API documentation](https://openfoodfacts.github.io/openfoodfacts-server/api/ref-cheatsheet/):

> "The future of search in Open Food Facts is the **search-a-licious** project, deployed, in beta, at `search.openfoodfacts.org`."

> "Full text search currently works only for v1 API (or search-a-licious, which is in beta)."

The legacy Perl-based search backend that powers both `/cgi/search.pl` and `/api/v2/search` is effectively deprecated. OFF's recommended replacement is their [Search-a-licious API](https://search.openfoodfacts.org/docs) (v0.1.0), an Elasticsearch-powered service.

kshoichi:
I just found this app and it'll be everything I'm looking for when the food search is fixed. Barcode entry is working. Text entry is broken (as described), however mine has been able to pull food details twice in the last 24 hours. All other attempts failed (probably a couple dozen tries).
I'm looking forward to using this when I can add my meals!

apedley:
Are you guys still having issues? Open Food Facts was having issues over the weekend but now I'm able to search by name almost everytime in the app without retrying. From what I could gather, the OFF Search a licious API will return less nutrient information so it makes sense to stay on the old api for as long as possible.

@kshoichi I'd definitely recommend setting up at least USDA as an external provider in settings. OFF data is from the public so it can vary in quality.

kshoichi:
Yes, OFF is pretty hit or miss for me. Sometimes it just errors out with a wall of debug text and no results. Sometimes I get the debug text and the results also come through. I went directly to their site and tried searching there and got "Our services are currently experiencing unusually high demand, or the page you requested is not available to anonymous users. We apologize for any inconvenience." So they have some work to do on their end. 

Thanks for the tip on the USDA - that's working perfectly so far so I'll stick with that for now! 

[...]

PR #1607, opened 2026-06-22
feat(food-search): unified single-field food search (Phase 1)

### LINKED ISSUE #1576: [Feature]: Search includes Foods & Meals

Sometimes I forget if I've entered a custom food (when I know all the macros) or entered each ingredient as a meal, and need to search "Database" and "Meals"

Could the search default to showing results for Foods and Meals?

I appreciate having the online search separate.

_No response_

Nice to have

_No response_

### Additional context

_No response_

jsandai:
## Description

Logging a food currently makes the user pre-decide where it lives (the Search, Online, or Meals tab) before seeing any results. This is Phase 1 of the unified search from #1496, and it delivers the foods-plus-meals merge asked for in #1576: one search field that returns local foods and saved meals together, with the default online provider streamed in below.

Client-side only, no backend or endpoint changes. The Search / Online / Meals tab control is removed. Typing runs the existing local-food, saved-meal, and default-provider searches together and renders them as sectioned results (Your Foods, Your Meals, then the provider). Web reuses the existing result card and its food/meal/provider badges. Mobile adds a small anchored New Food / New Meal menu in place of the tab-aware '+' button. Online search auto-runs, debounced, from 3 characters, using the user's default provider, and the local query is debounced too. The meal and online searches guard against stale resolves overwriting newer results.

Out of scope for this PR (later phases of #1496, per the #1576 issue discussion): the All Providers aggregation screen, the auto online search when there is no local match, and web auto-search-on-type refinements.

Linked Issue: Closes #1576 and partially addresses #1496

## How to Test

Web:
1. Open Add Food on a meal in the Diary. Confirm there is one search field and no Search/Online/Meals tabs.
2. Type a term that matches both a saved food and a saved meal. Confirm a Your Foods section and a Your Meals section appear together, and the default online provider's results appear below.
3. Confirm Custom Food, Import from CSV, and Scan Barcode still work, and the provider selector switches the online section.
4. On the Foods page import flow (online-only), confirm only online results show (no local foods or meals).

Mobile:
1. Open Add Food. Confirm one search field, no segmented tabs.
2. Type a matching term. Confirm Your Foods and Your Meals sections render together, with the default provider below.
3. Tap the plus button. Confirm a small anchored New Food / New Meal menu appears at the button (not a bottom sheet). In meal-builder mode, confirm meals are never shown and the plus creates a food directly.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

No backend changes in this PR (no migration, no new endpoint), so the backend Code Quality and Database Security items do not apply.

## Screenshots

<details>

### Before

web:

search-old

mobile:

Screenshot_20260622_103035_SparkyFitness

### After

web:

search-new

mobile:

Screenshot_20260622_103013_SparkyFitness

</details>

## Notes for Reviewers

This is Phase 1 of the redesign discussed on #1576, scoped to the single-field sectioned search. The All Providers screen and the online-fallback / web auto-search items are intentionally left for the follow-up phases per @apedley's request.

One deliberate web/mobile asymmetry to flag: the plus menu offers New Food and New Meal on mobile, but web offers New Food (Custom Food) only. Web meal creation lives in a separate MealBuilder flow that does not embed cleanly in this dialog, so I left New Meal off web for now rather than force it. Happy to wire it in a follow-up if you want strict parity.

Existing embedding modes are preserved: hideDatabaseTab now means online-only (the Foods page import flow) and hideMealTab hides the meals section (meal builder). `pnpm run validate` (typecheck + lint + format) passes; new web strings were added to the English translation file.


jsandai:
@gemini-code-assist

Good catch, addressed. Removed the now-dead hasUserSelectedProvider ref and simplified the effect to just sync the selected provider to the user's default (or first available) as providers load, since there's no manual provider selector on this screen anymore.

jsandai:
@gemini-code-assist

I think the current form is safe to leave as-is. `providers` comes from a React Query hook, which does structural sharing by default, so its reference stays stable across renders and same-data refetches (it only changes when the provider list actually changes). And even if the effect re-runs, `setSelectedProvider` is called with the default/first provider id, which is idempotent: setting state to the identical value is a no-op and triggers no re-render, so there's no extra re-sync to optimize away. Happy to add an explicit stable-id guard if you'd prefer the intent spelled out, but it wouldn't change runtime behavior.

jsandai:
Also worth noting:

The CI test failures in FoodSearchScreen.test.tsx are expected - they're testing the old tabbed UI that this PR intentionally removes. With the unified single-field search design, three tests are now obsolete:

1. "keeps meal rows as quick-log picks" - tries to find "Estimate from photo" CTA (line 202), which was removed as part of the unified search redesign
2. "renders the server error message when the search hook surfaces one" - tries to click the "Online" tab (line 251), which no longer exists
3. "toasts the error but still opens partial info when a detail fetch fails" - tries to click the "Online" tab (line 313), which no longer exists

Recommend deleting these three test cases from __tests__/screens/FoodSearchScreen.test.tsx since they're testing discarded UI behavior. The unified search behavior should be covered by new tests that:

1. Verify local foods, local meals, and provider results render together in sections
2. Confirm online results stream in automatically without a manual tab switch
3. Validate the new "New Food / New Meal" menu replaces the tab-aware add button (mobile)

The existing test mocks are still valid and reusable for these new scenarios.

CodeWithCJ:
@apedley  could you review this.

apedley:
1. This part of the user flow I outlined is missing on mobile and incorrect on web:

> The user can tap a button/link to change the provider shown but the user taps another option. "Show All Providers"

- Mobile has no way to view another providers results. You can only see the default provider. 
  - Make the header tappable (with affordance) and open a context menu or bottom sheet for selection
- Web allows you to change providers, but it also changes your default provider at the same time. This is supposed to allow the user to peak at another providers results if they don't see what they are looking for.

2. Anchor menu appears in different places on ios and android. 

ios which I believe is correct:

simulator_screenshot_22DECDC1-42D2-4BCA-9C9B-2004E0F15949

android:

image

apedley:
Also go ahead and fix/delete tests if they don't make sense anymore.

jsandai:
> 1. This part of the user flow I outlined is missing on mobile and incorrect on web:
> 
> > The user can tap a button/link to change the provider shown but the user taps another option. "Show All Providers"
> 
> - Mobile has no way to view another providers results. You can only see the default provider. 
>   - Make the header tappable (with affordance) and open a context menu or bottom sheet for selection
> - Web allows you to change providers, but it also changes your default provider at the same time. This is supposed to allow the user to peak at another providers results if they don't see what they are looking for.
> 
> 2. Anchor menu appears in different places on ios and android. 
> 
> ios which I believe is correct:
> 
> simulator_screenshot_22DECDC1-42D2-4BCA-9C9B-2004E0F15949
> 
> android:
> 
> image
> 

Understood - working on it but I have an errand to run. I'll ping you when it's done

jsandai:
@gemini-code-assist. Notes on each item:

**Mobile debounce (high):** already handled. `useFoodSearch` and `useMealSearch` debounce internally before any query runs (`const debouncedSearch = useDebounce(searchText.trim(), 300)`), and only fetch once the debounced term is >= 2 characters. So passing `searchText` straight in does not query on every keystroke; the keystrokes are absorbed inside the hook. No change needed here.

**Online loading state during the debounce window (medium):** good catch, fixed. `isOnlineLoading` was only set inside the 600ms timeout, while the local debounce settles at 300ms, leaving a short gap where the previous provider's results lingered as if they belonged to the new term. The loading state is now set as soon as the term changes, before the debounce timer. The existing `< 3 character` and no-provider guards still prevent a spurious spinner.

**`statusBarTranslucent` on the anchored menu Modal (medium):** leaving this as the explicit `StatusBar.currentHeight` offset rather than `statusBarTranslucent`. We actually tried the translucent-Modal approach in an earlier round; it shifted the menu out of alignment on a maintainer's physical Android device, which is what prompted the move to a manual offset. The current approach has been confirmed on a physical Android device with the trigger fully visible above the menu, matching iOS. Safe-area insets were also avoided here because they read unreliably (often zero) inside a React Native `Modal`, which is a separate native window. Happy to revisit if it misbehaves on any specific device.

jsandai:
@apedley - Both items are addressed, and I reworked the mobile tests per your note. Details below.

### 1. Provider visibility and the default-change bug

Mobile previously had no way to see any provider other than the default. The online section header is now the affordance: it is a tappable `BottomSheetPicker`, so the user can switch the provider they are previewing. Web previously changed the user's default provider at the same time as showing another provider's results; switching is now a peek only. The provider `Select` sets a local override (`setManualProviderId`) and no longer persists the default, so glancing at another provider does not silently re-point it. Verified on the emulator by switching between providers (for example Swiss Food Database to Open Food Facts).

Related UX touch on the same control: tapping the provider header now calls `Keyboard.dismiss()` first, so the bottom sheet animates up over a dismissed keyboard instead of opening behind it.

### 2. Anchored menu position differing between iOS and Android

The New Food / New Meal menu is positioned by measuring the trigger with `measureInWindow` and rendering the menu in a `Modal` overlay. Under Android edge-to-edge (enforced on Expo SDK 55) the trigger is measured in the app content window, which sits below the status bar, while the Modal overlay renders in screen-absolute space. The fixed gap between the menu and the trigger could not absorb that status-bar delta, so on devices with a taller status bar than the test emulator the menu rode up and clipped the trigger.

The fix adds `StatusBar.currentHeight` back on Android, so the menu lands just under the trigger on any device regardless of status-bar height. iOS reports no `currentHeight` and is unchanged, which keeps it matching the screenshot you posted. I used `StatusBar.currentHeight` rather than safe-area insets here on purpose, because safe-area insets are unreliable inside a React Native `Modal` (it is a separate native window and the insets frequently read as zero). Confirmed on a physical Android device that the trigger is now fully visible above the menu.

### Mobile test reconfiguration

I rewrote `FoodSearchScreen.test.tsx` rather than patching it, because several tests asserted behaviour the unified search intentionally removes, so keeping them green would have meant testing a UI that no longer exists.

Removed:

- The Meals-tab quick-log assertion (meals are a section now, not a tab).
- The "Estimate from photo" empty-state CTA tests, including the meal-builder hide case (that CTA and its deep link are not part of this flow).
- The two online-search error-message tests that asserted against the old tabbed online view's inline error rendering.

Added:

- Renders local foods, saved meals, and the online provider together in one search.
- Opens FoodEntryAdd when a saved-meal result is tapped.
- Does not show saved meals in meal-builder mode.
- Toasts the error but still opens partial info when an online detail fetch fails (the one error path still meaningful under the new flow, rewritten against it).

The Mobile, Frontend, and Server test jobs are passing on the latest commit. 

I can adjust any of this if you'd prefer a different approach, and if you could verify the menu is still good on iOS I'd appreciate it.

Screenshot_20260622_205454_SparkyFitness

Screenshot_20260622_205500_SparkyFitness

Screenshot_20260622_205515_SparkyFitness

---

One thing worth flagging that surfaced while testing this on a dev environment, though I believe it is pre-existing and separate from this PR:

Online search 500s for non-USDA providers (confirmed on Open Food Facts) with `Internal response validation failed`. The cause is in the v2 search response validation rather than this PR's UI. `openFoodFactsService` builds its pagination from the Open Food Facts `cgi/search.pl` JSON, which returns `page`, `page_size`, and `count` as strings, and passes them straight through (`page: data.page || page`). `SearchResponseSchema`/`PaginationSchema` then expect `z.number()`, so the parse throws. USDA passes because its adapter builds numeric pagination. The failing files (`integrations/openfoodfacts/openFoodFactsService.ts`, `routes/v2/foodRoutes.ts`) are not touched by this PR, and the old tabbed UI hits the same path, so this is not a regression introduced here. The unified search just makes it more visible, since it queries the default provider automatically instead of waiting for an Online tab click.

The fix is small either way. The most defensive option is coercing the pagination fields at the schema (`z.coerce.number()`), which covers any provider that returns string pagination. The narrower option is coercing at the source in the Open Food Facts adapter (`Number(data.page)` and friends).

Would you prefer this folded into this PR as a separate commit, or kept as a separate change so this one stays focused on the search UX? Happy to do either.

jsandai:
I went ahead and created a separate issue for the bug mentioned above - #1619

Figured separation of concerns is the cleaner route.

apedley:
On web, barcode scan results are hidden from view when the search field is empty
For both mobile and web, when you swap providers it will change the header to say the new provider before the results have loaded. Small thing but it can make it seem like swapping didn't work on a slow connection.
On mobile, the header to switch providers needs affordance.. it technically looks tappable but it sorta looks like it will collapse the section instead of opening a picker. Maybe change it so it looks like

`External Results						Source: USDA \/`

That looks a little more tappable and gives it a good section heading too.

jsandai:
Thanks Andrew, all three are addressed and pushed. Screenshots below.

1. Barcode results with an empty field (web). The scanned product was being set in state but hidden behind the search-mode gate, so it never rendered with an empty query. A barcode scan now keeps its result in its own state and shows it under a "Scanned product" section regardless of the search field, and it is cleared once you start a new search.

2026-06-23 22_45_30-Greenshot

2. Header changing before results load (web and mobile). Switching the provider now clears the previous provider's results so only a loading spinner shows under the new header, instead of stale rows that read as if nothing happened. On mobile the chevron in the header also becomes a spinner, and results still persist smoothly while typing within a provider (only a provider swap clears them).

2026-06-23 22_46_23-Greenshot

Screenshot_20260623_231704_SparkyFitness

3. Mobile header affordance. The online section header now reads "External Results" on the left with a tappable "Source: <provider>" control on the right (with the chevron), per your suggestion, so it reads as a source picker rather than a section collapse. Tapping it opens the provider bottom sheet.

Screenshot_20260623_220817_SparkyFitness

Let me know if you would like anything else adjusted.

apedley:
The spinner is a nice touch. Sorry I'm being picky but can we lose the outline? I've been trying to avoid outlines when they put a box in a box like that.

jsandai:
Done. I should have checked before building that - it now matches the apps existing pickers for consistency.

Screenshot_20260624_000615_SparkyFitness

Again, let me know if you want any other changes.

apedley:
The discoverability is still too low. I feel like the control being inside of the section header works against it. What do you think about something like this?

simulator_screenshot_83E37BEC-C5D8-4E09-82A9-85F7C316777A

jsandai:
I agree, even with the chevron there it's not obvious that the user can switch providers. Your color differentiator helps - I'm unsure what value the source label adds in this scenario; I'm thinking we could drop that and keep everything on one line?

apedley:
You're right I was worried about people missing it but the color is probably enough. I originally had it all left aligned which makes a bit more sense and is easier to scan because putting buttons in section headers is kind of an antipattern. This screen is desperate for free space though so it's defensible to do it. The color and icon being the double arrow should be enough.

jsandai:
Screenshot_20260624_144333_SparkyFitness

apedley:
Thanks for taking care of that. I think we are good here. Do you want to tackle phase two or should I?

jsandai:
Glad phase 1 is in good shape, thanks for the thorough review on it.

Happy to take phase 2 on, but I want to flag a couple of things up front since it is a bigger and fuzzier piece than phase 1, and I would rather align before anyone writes code.

The part I would want to nail down first is the "Top Matches" ranking and dedup. Phase 1 was mostly reorganising results that already existed; the All Providers view introduces a genuinely new problem, which is ordering the same food across heterogeneous providers (the Big Mac from FatSecret vs USDA vs OFF) and collapsing near-duplicates into one sensible top list. There is no obvious right answer there, the providers do not return comparable relevance scores, and the names and serving sizes differ, so I expect that is where most of the iteration will live rather than in the UI. How are you picturing it? Even a rough rule (for example: interleave one from each provider by name-match strength, dedupe on normalised name + brand) would give me something to build to.

The other thing worth flagging from phase 1 testing is rate limits, which I don't think are purely theoretical. OFF's legacy cgi/search.pl endpoint already 503s for us intermittently (that is the #1619 / #1079 thread), and fanning out to every active provider on each search will hit that harder, plus FatSecret and USDA have their own quotas. So whoever builds it, the All Providers path probably needs explicit partial-result handling (render whatever returns, skip the ones that fail) and maybe light caching or backoff, rather than assuming every provider answers. Not a blocker, just something to design in from the start instead of bolting on later.

If we do run with it, the things I would want locked before starting are: the ranking and dedup approach for Top Matches, the exact All Providers trigger and layout (your mock covers most of this already), the Show all screen behaviour (pushed per-provider screen, paginated, reusing the single-provider search), and the partial-failure or rate-limit strategy. With those settled it is very doable.

That said, if you would rather own phase 2 yourself, that is completely fine and probably sensible given it is your UX vision and your call on the ranking. Either way I am happy to hop on Discord to hash out the ranking and the provider fan-out before any code gets written, whichever of us ends up holding it.

apedley:
I honestly wasn't planning on any deduplication. If the user taps an "all providers" link they should expect to sees the same item multiple times. I'm thinking we could do 1 item from each provider (maybe 1 to 2 for a total of 5?) interleaved under "Top Matches" then go into each provider. Here's a quick mockup that's based on our current search screen (don't worry about following this too closely, use your best judgment)

image

What do you think for pagination? I was imaging no pagination and instead have a link that opens yet another search screen that searches just that provider.

For rate limits, searching all providers at once doesn't change the amount of api calls per service compared to normal searching so I don't think it will be a problem.  I think we need partial results handling anyway since OFF is flaky.

jsandai:
This is great, your answers knock out the part I was most worried about. Dropping dedup and treating Top Matches as a simple interleave (one or two per provider, around five, each tagged with its source) removes the fuzzy ranking problem entirely, which was the bulk of my concern.

From your mockup, here is how I am reading the plan, tell me if I have it right:

- Top Matches: a few interleaved results across providers, each row badged with its source.
- By Source: one expandable section per active provider, each with a result count badge, showing a handful of results, with the providers loading independently so they stream in (the USDA spinner and skeletons in your mock).
- Show all: a per-provider link that pushes a single-provider search screen, reusing the existing search rather than paginating inline.
- Partial results: render each provider as it returns and skip any that fail, which also covers OFF being flaky.

Two small things to confirm so I build the right thing:

1. For the By Source groups, is tapping the provider header meant to expand the section inline (the accordion in your mock), with a separate "show all N" link to the full single-provider screen? Or should the provider row go straight to the full screen with no inline expand? Either is easy, I just want to match what you are picturing.
2. Fair point on rate limits, each service still gets one call per search either way. I would run the fan-out client-side (parallel per-provider searches reusing the existing endpoint) so there is no new aggregation backend, and lean on the partial-results handling for the flaky ones. Shout if you would rather it be a single server-side call.

If that lines up I am happy to take phase 2 and run with it.

apedley:
> For the By Source groups, is tapping the provider header meant to expand the section inline (the accordion in your mock), with a separate "show all N" link to the full single-provider screen? Or should the provider row go straight to the full screen with no inline expand? Either is easy, I just want to match what you are picturing.

I was thinking tapping the header would expand the accordion with a separate "show all N" link. Just a fair warning though I might change my mind when I see the UX in action.

> I would run the fan-out client-side (parallel per-provider searches reusing the existing endpoint) so there is no new aggregation backend

Sounds good

> If that lines up I am happy to take phase 2 and run with it.

Great. I'm working on the chatbot stuff so this is all yours.

CodeWithCJ:
@jsandai  due to other PR. this now has conflcit. please take care ofthis during your next commit.


```
SparkyFitnessMobile/src/screens/FoodSearchScreen.tsx
```

jsandai:
Thanks for the heads up @CodeWithCJ.

Conflict resolved and pushed. Quick summary of what it was and the call I made, since it touches @thorethy1's native-header work and is worth your sign-off.

The only conflict was `SparkyFitnessMobile/src/screens/FoodSearchScreen.tsx`, and it was not really mechanical. #1515 added the native iOS header (Cancel / Add in the nav bar) to that screen, but this PR rewrites the same screen for the unified search, removing the structure #1515's header was built on (the `activeTab` state, the `SegmentedControl`, and the `activeTab`-based trailing action). The one genuine incompatibility is the "+" button: the unified search opens an anchored popover (New Food / New Meal) positioned from the on-screen button, which cannot live in a native nav-bar item.

I resolved it in favour of this PR's rewrite, so FoodSearchScreen keeps the unified search's own header and #1515's native-header treatment is not applied to this one screen for now. Everything else merged cleanly, and the branch is mergeable again.

Re-applying the native header to the rewritten screen is a small, self-contained follow-up. On iOS it would mean a native Cancel/Add nav bar plus an action sheet for the "+" (New Food / New Meal), since the popover cannot anchor to a nav-bar button; Android keeps the current popover. I can to do that either now in this PR or as a fast follow right after merge, whichever you and @apedley prefer. If you would rather keep #1515's screens fully consistent before this merges, say the word and I will fold the native-header integration into this PR. 

I also have phase 2 of this unified search ready to PR/review but obviously phase 1 needs to be in place first. Just an FYI - the All Providers follow-up (phase 2) extends this same screen, so it will hit and be resolved the same way when it comes through.

eta: that GitGuardian hit is the CI test DB password (postgres / sparky_app_pw) in .github/workflows/ci-tests.yml — it's already in main and our branch never touched that file; the merge commit just re-surfaced it on the PR. Safe to dismiss/merge through.

apedley:
> The one genuine incompatibility is the "+" button: the unified search opens an anchored popover (New Food / New Meal) positioned from the on-screen button, which cannot live in a native nav-bar item.
> 
> I resolved it in favour of this PR's rewrite, so FoodSearchScreen keeps the unified search's own header and #1515's native-header treatment is not applied to this one screen for now. Everything else merged cleanly, and the branch is mergeable again.


That's the right call and you can leave it at this for now. We won't have another release until after this feature is complete so I'm not worried about it shipping half done. After phase 2/3 we can revisit this header.

CodeWithCJ:
@jsandai  @apedley  is this ready to merge?

apedley:
> @jsandai @apedley is this ready to merge?

yes it's ready

CodeWithCJ (approved):

PR #1620, opened 2026-06-23
fix: coerce string-typed pagination fields in food search response (#1619)

### LINKED ISSUE #1619: [Bug]: Online food search 500s for providers that return string pagination (e.g. Open Food Facts)
### What happened?

Online food search returns a 500 for providers whose search API reports pagination fields as strings. Open Food Facts is the clearest case: searching it returns `500 Internal response validation failed`, and the client shows an `API Error: Internal response validation failed` toast with no results. USDA and other providers that return numeric pagination work normally, so the failure is provider-specific rather than a general search outage.

The root cause is response validation, not the food data. The v2 search route validates its response with `SearchResponseSchema` (`routes/v2/foodRoutes.ts`), whose `PaginationSchema` declares `page`, `pageSize`, and `totalCount` as `z.number()`. The Open Food Facts adapter builds pagination straight from the `cgi/search.pl` JSON (`integrations/openfoodfacts/openFoodFactsService.ts`: `page: data.page || page`), and that endpoint returns `page` as a string at least some of the time, so the parse throws and is mapped to a 500. The field type is inconsistent on Open Food Facts side (observed both as a string and as a number across requests), which is why the failure can look intermittent.

### Steps to reproduce

1. Configure Open Food Facts as a food data provider and set it as the default (or select it in search).
2. Open the food search and search for a common term, for example `yam`.
3. Observe the request to `GET /api/v2/foods/search/openfoodfacts?query=yam` return 500 and the `Internal response validation failed` toast.
4. Repeat with USDA selected and note it returns results normally.

### Expected behavior

Provider search should return results for Open Food Facts (and any provider that reports pagination values as strings) rather than failing response validation. String-typed numeric pagination fields should be accepted and normalized to numbers.

Docker Compose

### SparkyFitness version

v0.17.2

### Which area is affected?

Server / API, Integrations (Garmin, Fitbit, etc.)

```yaml
Not configuration-specific. Reproduces on a standard deployment with Open Food Facts configured as a provider.
```

```shell
[ERROR] v2 search response validation failed: ZodError: [
  {
    "expected": "number",
    "code": "invalid_type",
    "path": [ "pagination", "page" ],
    "message": "Invalid input: expected number, received string"
  }
]
    at searchHandler (/app/SparkyFitnessServer/routes/v2/foodRoutes.ts:217)
[ERROR] Error caught by centralized handler: Internal response validation failed
```

Image

### Anything else?

Why this can be easy to miss, and who is exposed:

The field type is inconsistent on the Open Food Facts side. The same `page=1` request was observed returning `page` as a string in one call and as a number in another, so the failure is effectively per-request rather than a stable condition, which makes it intermittent and hard to reproduce on demand. Exposure then scales with how often Open Food Facts is actually queried. In the current tabbed search UI, online provider search only runs when the user explicitly opens the Online tab, so most searches (local database) never touch a provider and the intermittent failure is rarely encountered. A search flow that queries the default provider automatically (rather than behind an Online tab) hits Open Food Facts far more frequently and surfaces the intermittent failure consistently.

Blast radius is broad because of the default seeding. `create_default_external_data_providers()` (run by the user-creation trigger for new users, and backfilled to existing users) seeds Open Food Facts as an active provider on every installation, and it is one of only two food providers seeded by default (Swiss Food Database is the other; USDA, FatSecret, etc. are user-added). With the frontend falling back to the first/active provider when no default is set, a typical installation has Open Food Facts selected without the user configuring anything.

Minimal fix identified and tested: coerce the numeric pagination fields in `PaginationSchema` (`SparkyFitnessServer/schemas/foodSchemas.ts`) with `z.coerce.number()`, leaving `hasMore` a strict boolean. This accepts string-typed values from any provider and does not regress providers that already return numbers (verified against the schema directly, and end to end on a dev deployment where Open Food Facts search returns 200 with valid pagination after the change). Coercing at the schema is preferred over patching a single adapter because it covers every provider defensively.

Relationship to existing reports: this is distinct from #1079 (the legacy `cgi/search.pl` endpoint returning HTTP 503 globally), which was closed as completed once that outage resolved. The migration to the Search-a-licious API proposed in #1479 was closed without merging, so the legacy endpoint is still in use, and with it this response-validation failure on string pagination. I did not find an existing open or closed issue covering this specific validation bug.

Separately, that same legacy endpoint still intermittently returns an HTML error page rather than JSON, which also surfaces as a 500 (the behavior originally reported in #1079). That is independent of this validation bug and could be handled as graceful degradation, or resolved outright by reviving the #1479 endpoint migration, in a follow-up.

MohammadYusif:
> [!TIP]

## Description

Online food search returns a 500 (`Internal response validation failed`) for providers that report pagination fields as strings — most visibly Open Food Facts, which is seeded as an active provider on every install. This drops all results for those providers.

`PaginationSchema` in `SparkyFitnessServer/schemas/foodSchemas.ts` declared `page`/`pageSize`/`totalCount` as `z.number()`. Open Food Facts' legacy `cgi/search.pl` endpoint returns those as strings (inconsistently — the same field flips between string and number across requests), so `SearchResponseSchema.parse()` in the v2 food search route threw a `ZodError` that the route mapped to a 500. The fix coerces those three numeric pagination fields with `z.coerce.number()` at the schema boundary (`hasMore` stays a strict boolean). Coercing at the schema rather than in a single adapter normalizes string-typed pagination defensively for every provider, and does not regress providers that already return numbers.

Linked Issue: Closes #1619

## How to Test

1. Check out this branch, then in `SparkyFitnessServer/` run `pnpm install`.
2. Run the new regression test: `pnpm exec vitest run tests/foodSchemas.test.ts` — 4 tests pass.
3. To confirm the test exercises the fix, temporarily revert the three `z.coerce.number()` calls back to `z.number()` in `schemas/foodSchemas.ts` and re-run: the two string-pagination cases fail with the same `invalid_type` ZodError seen in the issue.
4. End to end: with Open Food Facts selected as the provider, search a common term (e.g. `yam`) against `GET /api/v2/foods/search/openfoodfacts` and confirm it returns 200 with results instead of the `Internal response validation failed` toast.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

N/A — backend-only change, no UI.

## Notes for Reviewers

- Scope is intentionally minimal: the schema change plus a regression test. No new tables, so `rls_policies.sql` is unchanged (Database Security box left unchecked as N/A).
- `pnpm exec eslint schemas/foodSchemas.ts tests/foodSchemas.test.ts` passes clean, and `prettier --check` passes for both files.
- `pnpm run typecheck` currently reports pre-existing errors on `main` (e.g. in `auth.ts` and `tests/exerciseService.groupedWorkouts.test.ts`) that are unrelated to this change; this PR introduces no new typecheck errors in the touched files.
- I considered normalizing in the Open Food Facts adapter instead, but coercing at `PaginationSchema` covers any provider that reports string pagination, which the issue notes is a recurring shape. Happy to move it into the adapter if you'd prefer per-provider handling.

### DIFF
--- a/SparkyFitnessServer/schemas/foodSchemas.ts
+++ b/SparkyFitnessServer/schemas/foodSchemas.ts
@@ -58,10 +58,11 @@ export const PaginationSchema = z.object({
   // these pagination values as strings, and the same field can switch between a
   // string and a number across requests. Coerce them so any provider that sends
   // string-typed numeric pagination is normalized rather than failing response
-  // validation. `hasMore` stays a strict boolean.
-  page: z.coerce.number(),
-  pageSize: z.coerce.number(),
-  totalCount: z.coerce.number(),
+  // validation. `.int()` makes the integer intent explicit and rejects
+  // non-integer floats (e.g. "1.5"). `hasMore` stays a strict boolean.
+  page: z.coerce.number().int(),
+  pageSize: z.coerce.number().int(),
+  totalCount: z.coerce.number().int(),
   hasMore: z.boolean(),
 });
 
--- a/SparkyFitnessServer/schemas/foodSchemas.ts
+++ b/SparkyFitnessServer/schemas/foodSchemas.ts
@@ -54,9 +54,14 @@ export const NormalizedFoodSchema = z.object({
 export type NormalizedFood = z.infer<typeof NormalizedFoodSchema>;
 
 export const PaginationSchema = z.object({
-  page: z.number(),
-  pageSize: z.number(),
-  totalCount: z.number(),
+  // Some providers (e.g. Open Food Facts' legacy cgi/search.pl endpoint) report
+  // these pagination values as strings, and the same field can switch between a
+  // string and a number across requests. Coerce them so any provider that sends
+  // string-typed numeric pagination is normalized rather than failing response
+  // validation. `hasMore` stays a strict boolean.
+  page: z.coerce.number(),
+  pageSize: z.coerce.number(),
+  totalCount: z.coerce.number(),
   hasMore: z.boolean(),
 });
 
--- a/SparkyFitnessServer/tests/foodSchemas.test.ts
+++ b/SparkyFitnessServer/tests/foodSchemas.test.ts
@@ -1,0 +1,89 @@
+import { describe, expect, it } from 'vitest';
+import {
+  PaginationSchema,
+  SearchResponseSchema,
+} from '../schemas/foodSchemas.js';
+
+const validVariant = {
+  serving_size: 100,
+  serving_unit: 'g',
+  calories: 50,
+  protein: 1,
+  carbs: 10,
+  fat: 0,
+  is_default: true,
+};
+
+const validFood = {
+  name: 'Yam',
+  brand: null,
+  is_custom: false,
+  default_variant: validVariant,
+};
+
+describe('PaginationSchema', () => {
+  it('accepts numeric pagination fields and keeps them as numbers', () => {
+    const result = PaginationSchema.safeParse({
+      page: 1,
+      pageSize: 20,
+      totalCount: 42,
+      hasMore: true,
+    });
+    expect(result.success).toBe(true);
+    if (result.success) {
+      expect(result.data.page).toBe(1);
+      expect(result.data.pageSize).toBe(20);
+      expect(result.data.totalCount).toBe(42);
+    }
+  });
+
+  it('coerces string-typed numeric pagination fields to numbers', () => {
+    // Open Food Facts' legacy cgi/search.pl endpoint returns page (and at times
+    // page_size/count) as strings, which previously threw a ZodError and
+    // surfaced as a 500 "Internal response validation failed".
+    const result = PaginationSchema.safeParse({
+      page: '1',
+      pageSize: '20',
+      totalCount: '42',
+      hasMore: false,
+    });
+    expect(result.success).toBe(true);
+    if (result.success) {
+      expect(result.data.page).toBe(1);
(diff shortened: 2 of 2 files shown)


jsandai:
Thanks for jumping on this, @MohammadYusif. This matches the root cause from #1619, and coercing the three numeric pagination fields at the schema boundary (with `hasMore` left strict) is exactly the right call. Nice to see a dedicated schema regression test too.

One piece of context from the issue that is worth keeping in view for prioritization: Open Food Facts is seeded as an active provider on every installation (`create_default_external_data_providers()` runs on user creation and was backfilled to existing users), and it is one of only two food providers seeded by default. So this is not an edge case limited to users who configured a niche provider; a typical default install has Open Food Facts selected and is exposed to the 500.

Optional: appending `.int()` to the three fields makes the integer intent explicit and rejects a non-integer string like `"1.5"`. It is genuinely optional here since `z.coerce.number()` already rejects `NaN` in zod v4 and pagination values are always integers, so the coercion you have resolves the bug as-is.

CodeWithCJ:
@MohammadYusif  Just checking if you will be able to apply the suggestions that @jsandai  added

MohammadYusif:
Applied @jsandai's suggestion — appended `.int()` to the three coercion calls so a non-integer float like `"1.5"` is now rejected rather than silently coerced to `1.5`. Well-formed string pagination values (`"1"`, `"20"`) still coerce correctly. Added a regression test for the `"1.5"` case.

CodeWithCJ (approved):

PR #1658, opened 2026-06-27
feat(openfoodfacts): support self-hosted OFF via provider settings

petermcneil:
## Description

**What problem does this PR solve?**
The Open Food Facts integration hardcoded `https://world.openfoodfacts.org` for the search, product-by-barcode, and session-login endpoints, so deployments running a self-hosted instance (or hit by rate limits) couldn't redirect lookups there.

**How did you implement the solution?**
Per the discussion on #1657, this no longer uses an env var. Instead it reuses the existing `external_data_providers.base_url` column — the same field already used by the `mealie`/`tandoor`/`norish` providers — for `provider_type = 'openfoodfacts'` too, configurable per provider row in Settings, defaulting to the public `world.openfoodfacts.org` instance when unset.

- `openFoodFactsAuth.ts` gains a combined `resolveOpenFoodFactsProvider()` that resolves both the session cookie and the base URL from a single provider lookup (session-cookie auth and self-hosting are independent — a self-hosted provider with no login credentials still resolves its `base_url`).
- `openFoodFactsService.ts`'s search and barcode-lookup URLs are built from the resolved `base_url` instead of the hardcoded host.
- `getActiveOpenFoodFactsProviderId` / `resolveOpenFoodFactsProviderId` no longer require login credentials to select an OFF provider — previously a self-hosted-only provider (custom `base_url`, no login) could never be selected, since the seeded default OFF provider row has no credentials.
- Frontend: adds a "Base URL (Optional)" field to the OpenFoodFacts block in both the Add and Edit provider forms, with a translated label/help string.

No schema change, no new endpoint — `base_url` already exists on the table and is already covered by the existing RLS policy for the row.

Linked Issue: Closes #1657

## How to Test

1. Check out this branch and run `cd SparkyFitnessServer && pnpm install`
2. Run the test suite: `pnpm test` (2127 passed) — `tests/openFoodFactsService.test.ts`, `tests/openFoodFactsAuth.test.ts`, `tests/externalProviderService.test.ts`, and the new `tests/externalFoodSearchService.openFoodFacts.test.ts` cover the base_url resolution and credential-relaxation behaviour.
[...]

PR #1672, opened 2026-06-28
fix(openfoodfacts): switch to search-a-licious API with legacy fallback and optional auth

nicvoigt:
## Summary

- Replaces the single `cgi/search.pl` endpoint with a **dual-strategy** approach: the new [`search.openfoodfacts.org/search`](https://search.openfoodfacts.org) endpoint (search-a-licious) is tried first, and the legacy `world.openfoodfacts.org/cgi/search.pl` is used as a fallback. This fixes rate-limiting issues users were hitting with the older endpoint.
- Adds **optional authenticated sessions** for the legacy fallback via env vars (`SPARKY_FITNESS_OFF_USER_ID`, `SPARKY_FITNESS_OFF_PASSWORD`, `SPARKY_FITNESS_OFF_APP_*`), with a 30-minute session-cookie cache to avoid repeated logins.
- Makes both base URLs configurable (`SPARKY_FITNESS_OFF_WORLD_BASE_URL`, `SPARKY_FITNESS_OFF_SEARCH_BASE_URL`) for self-hosted setups.
- Fixes a vitamin C rounding precision issue in `mapOpenFoodFactsProduct`.
- Updates tests and `external_providers.md` docs to reflect the new strategy.

## Related

- Jira: KAN-539
- Similar scope to #1658 but goes further by using the dedicated search service rather than just making the legacy base URL configurable.

## Test plan

- [ ] Run `SparkyFitnessServer` unit tests: `npm test -- --testPathPattern="openFoodFacts"`
- [ ] Verify text search returns results for common food queries
- [ ] Verify barcode lookup still works (unchanged code path)
- [ ] With `SPARKY_FITNESS_OFF_USER_ID` + `SPARKY_FITNESS_OFF_PASSWORD` set, verify the legacy fallback authenticates and searches
- [ ] Verify graceful degradation when both endpoints fail

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/SparkyFitnessMobile/docs/external_providers.md
+++ b/SparkyFitnessMobile/docs/external_providers.md
@@ -12,6 +12,14 @@ They're all mounted under `/api/foods/` (since `foodIntegrationRoutes` is mounte
 
 Most providers require an `x-provider-id` header — that's the ID of the user's configured external provider record (from `/api/external-providers/user/:userId`). The middleware uses it to look up the stored API keys/credentials for that provider. Open Food Facts is the exception since it's a free public API.
 
+The server now prefers OFF's dedicated search service at `https://search.openfoodfacts.org/search` for text search because the older `cgi/search.pl` endpoint is rate-limited more aggressively. If you need authenticated OFF fallback for degraded scenarios, configure these server env vars:
+
+- `SPARKY_FITNESS_OFF_USER_ID`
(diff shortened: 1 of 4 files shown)


apedley:
The search a licious interface was providing much less information while being just as unstable the last time I checked.  Did they update things?

nicvoigt:
Tracking Jira ticket: KAN-539

PR #1695, opened 2026-07-01
feat(food-search): All Providers aggregated search on web (phase 3)

### LINKED ISSUE #1694: [Feature]: All Providers food search web parity (unified search phase 3)

Phase 2 added an "All Providers" mode to the mobile food search, so a single query fans out across every active provider at once. The web food search did not get that, so on web the online half of search still queries one provider at a time: to compare what each configured source has, you open the source switcher and step through providers one by one. This closes that gap so web and mobile behave the same.

An "All Providers" option in the web food search source switcher that searches every active food provider at once and shows the combined results in a single view:

- Top Matches: a short interleaved list pulling the top results from each provider round robin, each row tagged with a colour-coded source badge, and never fewer entries than the number of providers that returned results so every source is represented.
- By Source: one collapsible section per provider with a result-count badge, a per-provider loading state, an error state with retry, an empty state, and a "Show all N" link that drops into that provider's existing single-provider view.
- Each provider streams in independently, so a slow or failing provider does not hold up the others (partial results).

This keeps the "I just want to find the food, not pick where it comes from" goal of #1496 on web, while still letting a user narrow to one source when they want to.

Web frontend, Nutrition tracking

The phase 1 single-provider switcher already lets you look at one provider at a time on web. It works, but it does not let you see across providers in one pass, which is the gap this closes and which mobile already has.

### Additional context

Phase 3 of #1496: web parity of the phase 2 mobile feature (#1653). The implementing PR fans out client side (React Query useQueries, no new backend), reuses the same interleave and index-assigned colour logic as mobile, and includes unit tests plus screenshots of Top Matches, the By Source accordions, per-provider colours, and the loading/error/empty states.

jsandai:
## Description

Phase 3 of the unified search (#1496). Phase 2 added an "All Providers" mode to the mobile food search; this brings the same mode to the web `FoodSearch`, so web and mobile behave the same. When more than one food provider is active, a single search can fan out across every active provider at once instead of querying one at a time.

Selecting "All Providers" in the source switcher fans the search out across all active providers in parallel via a new declarative `useAllProvidersFoodSearch` hook (React Query `useQueries`), which runs alongside the existing single-provider search. Each provider is its own query, so results stream in independently and a slow or failing provider does not block the others (partial results). The view shows Top Matches (each provider's top results interleaved round robin, capped, but never below the number of providers that returned results so every source is represented, each row badged with a colour-coded source) and a By Source section with one collapsible accordion per provider (result-count badge, per-provider loading, error state with retry, empty state, and a "Show all N" link that deep-links into the existing single-provider view). Provider colours use an index-assigned palette (collision-free by list position) seeded from the app's existing chart colours. The fan-out is entirely client side, so there is no new backend. The interleave and colour helpers are shared pure functions with unit tests.

Linked Issue: Closes #1694

## How to Test

1. Have two or more active food providers configured (for example Open Food Facts plus another).
2. Open the food search and type a query (for example "salmon").
3. In the source switcher, choose "All Providers".
4. Verify Top Matches shows interleaved results badged by source, and By Source lists each provider with a result count.
5. Expand a provider; use "Show all N" to open that provider's full results.
6. Confirm a provider that errors or returns nothing shows its error (with retry) or empty state respectively, and the others still render.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

2026-07-01 17_16_49-_Search both meals and foods_ _ ccshuttle - Discord

</details>

## Notes for Reviewers

Web only; no `SparkyFitnessServer/` or `SparkyFitnessMobile/` changes (the mobile version shipped in phase 2). Tests: unit tests added for the `topMatches` interleave and `providerColor` assignment; the full frontend jest suite passes and typecheck is clean on the changed files.

Two deliberate choices worth surfacing, both easy to change:
- Provider colours reuse the web app's existing chart palette (static hex, index-assigned by list position, ordered so adjacent providers stay in clearly different hue families), rather than theme-aware CSS variables like the mobile version. This keeps them consistent with the report charts, but it means they do not adapt to light/dark. Happy to move them onto theme variables if you would prefer parity with the mobile approach.
- The mobile "cap the local Your Foods / Your Meals list while online results are showing" affordance is not included here, since the web layout presents local and online results differently. If the long-local-list-buries-online problem applies on web too, that is a small follow-up.

### DIFF
--- a/SparkyFitnessFrontend/src/utils/providerColor.ts
+++ b/SparkyFitnessFrontend/src/utils/providerColor.ts
@@ -34,29 +34,18 @@ export const PROVIDER_COLOR_PALETTE = [
 
 const FALLBACK_COLOR = '#94a3b8'; // slate-400, matches muted-foreground tone
 
-// Maps each provider id to a palette colour by list position. Collision-free
-// while the active providers fit the palette; wraps past that.
-export function buildProviderColorMap(
-  providers: DataProvider[],
-  palette: string[] = PROVIDER_COLOR_PALETTE
-): Map<string, string> {
-  const byId = new Map<string, string>();
-  if (palette.length > 0) {
-    providers.forEach((p, i) => {
-      byId.set(p.id, palette[i % palette.length] ?? FALLBACK_COLOR);
-    });
-  }
-  return byId;
-}
-
-// Returns a resolver mapping a provider id to its assigned colour. Build the map
-// once (e.g. via useMemo on the providers list) and call the returned function
-// while rendering rows.
+// Returns a resolver mapping a provider id to its assigned colour, by the
+// provider's position in the active list (palette[i % length]). Collision-free
+// while the active providers fit the palette; wraps past that. Build the
+// resolver once (e.g. via useMemo on the providers list) and call it while
+// rendering rows.
 export function makeProviderColorResolver(
-  providers: DataProvider[],
-  palette: string[] = PROVIDER_COLOR_PALETTE
+  providers: DataProvider[]
 ): (providerId?: string | null) => string {
-  const byId = buildProviderColorMap(providers, palette);
+  const byId = new Map<string, string>();
+  providers.forEach((p, i) => {
+    byId.set(p.id, PROVIDER_COLOR_PALETTE[i % PROVIDER_COLOR_PALETTE.length]);
+  });
   return (providerId?: string | null): string => {
     if (!providerId) return FALLBACK_COLOR;
     return byId.get(providerId) ?? FALLBACK_COLOR;
--- a/SparkyFitnessFrontend/src/utils/providerColor.ts
+++ b/SparkyFitnessFrontend/src/utils/providerColor.ts
@@ -44,7 +44,11 @@ export function makeProviderColorResolver(
 ): (providerId?: string | null) => string {
   const byId = new Map<string, string>();
   providers.forEach((p, i) => {
-    byId.set(p.id, PROVIDER_COLOR_PALETTE[i % PROVIDER_COLOR_PALETTE.length]);
+    byId.set(
+      p.id,
+      PROVIDER_COLOR_PALETTE[i % PROVIDER_COLOR_PALETTE.length] ??
+        FALLBACK_COLOR
+    );
   });
   return (providerId?: string | null): string => {
     if (!providerId) return FALLBACK_COLOR;
--- a/SparkyFitnessFrontend/src/utils/providerColor.ts
+++ b/SparkyFitnessFrontend/src/utils/providerColor.ts
@@ -1,0 +1,57 @@
+import type { DataProvider } from '@/types/settings';
+
+// Per-provider signature colours used to tell sources apart at a glance in the
+// "All Providers" aggregated search: a tinted badge behind Top Matches rows and
+// a dot before each By Source provider. Colours are assigned by the provider's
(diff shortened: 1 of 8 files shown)


jsandai:
Both applied.

The single-provider online-search effect now early-returns when All Providers is selected. You are right that it was re-running on every keystroke (searchTerm is a dependency), falling through to `setExternalResults([])`, and re-rendering the whole component each time since a fresh `[]` is never `===` the previous one. In All Providers mode the aggregated hook owns the online results, so the effect no-ops there now, and the aggregated flow already clears the single-provider results on switch, so it is safe.

Added `aria-expanded` to the By Source accordion buttons (`expandable ? expanded : undefined`) so screen readers announce the expanded/collapsed state.

`pnpm run validate` (typecheck, lint, prettier) passes and the food-search unit tests are green.

jsandai:
Applied. The v2 fan-out path now guards `data.foods` with `Array.isArray` before mapping, matching the Nutritionix guard, so a malformed payload (foods missing, null, or a non-array) falls back to an empty list instead of crashing the query. `pnpm run validate` passes.

/gemini review

jsandai:
@gemini-code-assist - looks like we're in edge case and polish territory so I'm calling it here. 

Both applied.

"Show all N" now resets the shared scroll container to the top when it switches to the single-provider view, so you land on the new results and loading spinner instead of being stranded at the bottom.

Added a hover background to the By Source accordion header (cleared when the row is disabled) for clearer interactive feedback.

`pnpm run validate` passes.

jsandai:
@apedley - this awaits your feedback/critique when you gat a chance

SparkyFitnessFrontend/src/utils/providerColor.ts
apedley, line 1: These functions take a palette but it's never actually sent. buildProviderColorMap is only used internally yet it's exported.

Can we trim this down some?

apedley (commented):
Looks good. Left one comment on the provider color.. I think it might be a little overengineered. One thing I noticed that's out of scope, currently only like 10 results are displayed when selecting a specific external provider. Was it like that before?

jsandai:
> currently only like 10 results are displayed when selecting a specific external provider. Was it like that before?

Yes, that is pre-existing and not something this PR changed. It is the food_display_limit user preference (defaults to 10), applied to single-provider online searches app-wide – the USDA/YAZIO handlers pass it straight through to the v2 search call. It is configurable in Settings, so it is not All-Providers-specific. Happy to raise the default or add pagination for the specific-provider view as a separate follow-up.

> These functions take a palette but it's never actually sent. buildProviderColorMap is only used internally yet it's exported.
> 
> Can we trim this down some?

Trimmed in c400668a. Dropped the palette parameter (every real caller used the default) and removed the internal-only buildProviderColorMap export, folding the map build into makeProviderColorResolver. providerColor is now just makeProviderColorResolver plus the PROVIDER_COLOR_PALETTE constant, and the tests exercise the public helper.

apedley:
Looks good to me

apedley:
> Happy to raise the default or add pagination for the specific-provider view as a separate follow-up.

We should probably add it as a follow up. If you could handle it that'd be great

CodeWithCJ (approved):

PR #1961, opened 2026-07-29
Support custom meal types in MCP food tools

### LINKED ISSUE #1959: [Bug]: MCP food tools cannot log entries to custom meal types
### What happened?

SparkyFitness supports custom meal types in the web application. For example, a user can create a separate custom meal type named `Second breakfast`, with its own `meal_type_id`, distinct from the default `Breakfast` and `Snacks` types.

However, the current MCP food write tool does not expose custom meal types:

- `sparky_manage_food` accepts only the fixed `meal_type` values `breakfast`, `lunch`, `dinner`, and `snacks`;
- it does not provide a `meal_type_id` argument;
- there is no MCP action for listing the user's available meal types and their IDs;
- `update_entry` cannot move an existing food entry to another meal type.

As a result, an MCP client cannot log food to a custom meal type. An AI assistant may incorrectly substitute `breakfast` or `snacks`, even though those are separate categories in the user's diary.

The regular application API already supports custom meal types and `meal_type_id`, so this appears to be a gap in the MCP schema rather than a limitation of SparkyFitness itself.

### Actual behavior

The MCP client must either:

- use an incorrect default category such as `breakfast` or `snacks`; or
- refuse to create the entry because the requested custom category cannot be represented.

This makes AI-assisted food logging unreliable for users who organize their diary with custom meal types.

### Additional context

This is not a request to treat `snacks` as an alias for `Second breakfast`. They are separate meal types with separate IDs in the same user account.

Related work, but not an exact duplicate:

- #507 added custom meal type support and changed the backend to use `meal_type_id`.
- #746 fixed forwarding `meal_type_id` in the food entry service.
- #1136 added changing an entry's meal type in the web interface.
- #1484 / #1485 expanded structured MCP food and diary access, but the current MCP food write schema still does not expose custom meal type IDs.

No private instance URL, authentication data, user IDs, entry IDs, or meal type UUIDs are included in this report.

### Steps to reproduce

1. In the SparkyFitness web interface, create a custom meal type named `Second breakfast`.
2. Verify that it appears as a separate diary category, distinct from `Breakfast` and `Snacks`.
3. Connect an MCP client and inspect the schema of `sparky_manage_food`.
4. Attempt to log a food entry to the custom `Second breakfast` category.
5. Observe that the tool only accepts the four fixed `meal_type` values and provides no way to pass the custom category's ID.
6. Log the entry under one of the fixed types and attempt to correct it with `update_entry`.
7. Observe that `update_entry` can change quantity/unit but not the meal type.

### Expected behavior

- provide a read-only MCP action/tool that lists available meal types, including `id`, `name`, and sort order;
- accept `meal_type_id` in food logging actions such as `log_food`, `log_external_food`, `create_food`, and `log_meal`;
- allow `update_entry` to change `meal_type_id`;
- return the resolved meal type ID and name when reading food diary entries;
- keep the existing fixed `meal_type` values as a backwards-compatible fallback.

If both fields are supplied, `meal_type_id` should take precedence.

Docker Compose

### SparkyFitness version

1.6.0

### Which area is affected?

Server / API

```yaml
- Deployment: Docker Compose, self-hosted
- MCP server: `sparkyfitness-mcp-server` 1.6.0
- MCP tool: `sparky_manage_food`
- Client tested: an MCP-compatible AI client
```

```shell

```

_No response_

### Anything else?

_No response_

Dragonk:
## Description

MCP food tools could only use four fixed meal types, while the REST API already supports custom meal types by ID.

Added meal type discovery, meal_type_id support with ID precedence and legacy fallback, entry moves, and resolved meal type data in diary reads. The shared foodEntryService keeps name-based resolution for existing REST/web/mobile clients; MCP passes the canonical ID through.

Linked Issue: Closes #1959

## How to Test

1. Run the targeted MCP tests: tests/chatbotToolSchemas.test.ts and tests/chatbotToolsFood.test.ts.
2. List meal types and log food using a custom meal_type_id.
3. Move an existing entry to a custom meal type and verify diary reads include its ID and name.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

- [x] **[MANDATORY for Backend changes] Code Quality**: Typecheck and full ESLint pass; the full server test suite passes (2498 tests, 203 files; targeted MCP + service tests included). Changed TypeScript files pass Prettier. The repository-wide Prettier check reports 552 pre-existing differences in untouched files.

## Notes for Reviewers

The PR changes the MCP food tool, its tests, MCP documentation, and a small shared-service helper (foodEntryService.resolveMealTypeId). No REST route or request-schema changes were made. The shared copy service now canonicalizes meal type selectors to IDs, which corrects ambiguous name-based behavior for existing REST/web/mobile clients.

## Summary by CodeRabbit

* **New Features**
  * Browse built-in and custom meal types, with hidden types excluded.
  * Select meal types by ID or name when logging, updating, copying entries, and saving templates.
  * View meal-type and entry identifiers in diary results.
* **Bug Fixes / Validation**
  * Support partial updates to quantity, unit, and meal type.
  * Improve validation and prioritize matching IDs over names.
  * Preserve omitted template quantity and unit values during updates.
* **Documentation**
  * Updated food management guidance for meal-type selection and listing.

---

### AI assistance disclosure

Parts of this implementation were drafted with Open WebUI.
All changes were manually reviewed, iteratively corrected, and validated by
Dragonk. The full server test suite, typecheck, lint, and formatting checks
were run before the final update.

### DIFF
--- a/SparkyFitnessServer/ai/tools/foodTools.ts
+++ b/SparkyFitnessServer/ai/tools/foodTools.ts
@@ -467,7 +467,7 @@ const DIARY_MEAL_DROP = [
 ] as const;
 
 interface ResolvedMealType {
-  id?: string;
+  id: string;
   name: string;
 }
 
@@ -483,7 +483,17 @@ async function resolveMealType(
     );
     return resolved ? { id: resolved.id, name: resolved.name } : null;
   }
-  return mealType ? { name: mealType } : null;
+  if (!mealType) {
+    return null;
+  }
+  // For built-in meal types (by name), resolve to actual ID
+  const mealTypes = await mealTypeRepository.getAllMealTypes(userId);
+  const normalizedName = mealType.trim().toLowerCase();
+  const resolved = mealTypes.find(
+    (type: { id: string; name: string }) =>
+      type.name.trim().toLowerCase() === normalizedName
+  );
+  return resolved ? { id: resolved.id, name: mealType.trim() } : null;
 }
 // Full food_entries dumps (`SELECT fe.*`, used by recent-entries and food-usage)
 // add audit/ownership columns on top of the diary projection's surrogate keys.
@@ -801,7 +811,7 @@ Actions:
 - update_entry(entry_id, entry_type, quantity?, unit?, meal_type_id?, meal_type?) — changes quantity/unit and/or moves the entry to another meal type.
 - update_food_variant(food_id?|variant_id?, serving_size?, serving_unit?, calories?, protein?, carbs?, fat?, saturated_fat?, fiber?, sugar?, sodium?, ..., update_existing_entries?) — updates an existing food variant without deleting the food. Defaults to leaving existing diary entries unchanged.
 - copy_from_yesterday(target_date?, source_date?, meal_type_id?|meal_type?)
-- save_as_meal_template(entry_date, meal_type_id?|meal_type?, meal_name, description?)
+- save_as_meal_template(entry_date, meal_type_id?|meal_type?, meal_name, description?) — REQUIRES EXPLICIT action field. Saves diary entries for a given date and meal type as a reusable meal template.
 - log_water(amount_ml, entry_date)
 - get_nutritional_summary(start_date, end_date) — returns macro breakdown for a range of dates
 - get_water_history(start_date?, end_date?)`,
@@ -839,6 +849,9 @@ Actions:
             if (args.food_name) {
               return 'lookup_food_nutrition';
             }
+            if (args.target_date || args.source_date) {
+              return 'copy_from_yesterday';
+            }
             if (args.meal_name && (args.meal_type || args.meal_type_id)) {
               return 'log_meal';
             }
@@ -863,23 +876,8 @@ Actions:
             if (args.food_id) {
               return 'delete_food';
             }
-            if (
-              args.target_date ||
-              args.source_date ||
-              args.meal_type ||
-              args.meal_type_id
-            ) {
-              return 'copy_from_yesterday';
-            }
             if (args.start_date || args.end_date) {
               return 'get_nutritional_summary';
-            }
-            if (
-              args.entry_date &&
-              args.meal_name &&
-              (args.meal_type || args.meal_type_id)
-            ) {
-              return 'save_as_meal_template';
             }
             if (args.entry_date) {
               return 'list_diary';
@@ -989,6 +987,7 @@ Actions:
           'log_food',
           'log_external_food',
           'log_meal',
+          'save_as_meal_template',
         ];
         if (
           mealTypeRequiredActions.includes(args.action) &&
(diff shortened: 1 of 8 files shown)


apedley (changes requested):
Hey thanks for taking care of this. The ID path stuff looks good and I appreciate the tests. There's one real bug, one small bug, and one area for expansion:

- bug: when a meal_type is pased without an id, it can say a tool was successful when it wasn't. The service will backfill `meal_type_id: entryData.meal_type_id ?? existingEntry.meal_type_id` (line 765) so we get the existing id, the name is only resolved when the id is absent so it skips this part and COALESCE on the meal_type_id will keep the old value.
- minor: list_meal_types will show hidden meal types, is_visible is currently dropped from the MCP projection so it's not considered
- expansion (i think we should update these too?): copy_from_yesterday and save_as_meal_template are still enum only for meal types

apedley (changes requested):
1. Sending meal_type_id as null alongside the name doesn't quite fix the problem. The service still has`meal_type_id: entryData.meal_type_id ?? existingEntry.meal_type_id`  (services/foodEntryService.ts services/foodEntryService.ts on 765). Sending `null` and the existing here will backfill the existing id. Then the reposiutory (models/foodEntry.ts) only resolves the name `if (!mealTypeId)` and the old ID still always wins through coalesce. The new regression test only tests that the function sent null so it doesn't cover this situation. Suggestion: change the service or the tool should resolve the name to an id itself 
2. list meal types is fixed
3. copy_from_yesterday and save_as_meal_template - the schema and service levels need updates as well. For copy, the schema is set to strict and never got a meal_type_id field so copy will fail with unrecognized_keys. And the service layer is bugged too, copyFoodEntries's target-side resolveMealTypeId by name only and it's sent a uuid.  For save, id gets dropped the same way and meal_type is still required in its schema. So save with an id only fails validation

CodeWithCJ:
@Dragonk CI Test for server is in failed state.

apedley:
Points 1 and 3 from before are still just as valid despite the latest commit message saying they are fixed. Things barely changed. My third point is causing the failing tests.

apedley (changes requested):
The three review items are addressed, thanks for fixing them. But there's now a failing test related to these changes in tests/chatService.test.ts. "executes a log_food tool call in-process…" fails with createFoodEntry: Number of calls: 0.

Fixing it needs two things: mock ../models/mealType in chatService.test.ts (like chatbotToolsFood.test.ts now does), and update the assertion at line ~478 — the payload now carries meal_type_id instead of meal_type: 'breakfast'. Please run the full pnpm test in SparkyFitnessServer/ before pushing.

apedley (approved):
Good job thanks for taking care of all that

Dragonk:
Thank you for your patience and thorough review. I am glad we successfully merged this PR.

PR #2031, opened 2026-08-03
Feature: Supplement tracking

### LINKED ISSUE #1769: [Feature]: Add Suplements/Vitamins

I dislike tracking vitamins or supplements in two places.

I know we can add vitamins and supplements to our food database. But with the new medication (and the reminders it offers) it would be really cool if we could add our supplements there, and still have it show up in our nutrient summary.

Nutrition tracking

Nice to have

_No response_

### Additional context

_No response_

jsandai:
## Description

There is nowhere sensible to track a supplement: logging a multivitamin as a food gives you no reminders or adherence, and logging it as a medication means its vitamin D never reaches your nutrition totals. This adds supplements as a subtype of the existing medication system, so one entry gets both.

A supplement is a medication row with two new columns: an `is_supplement` flag (mirroring `is_glp1`) and a per-dose `nutrients` JSONB payload. Logging a dose snapshots that payload onto `medication_entries.nutrients_snapshot`, exactly as the row already snapshots med name and dose, so history stays immutable when the medication is edited. The nutrition summary and the report each gained a supplement arm reading that snapshot.

On the Medications page, an `All | Meds | Supplements` filter partitions the existing list, the Log view renders a card per subtype so the two read as separate sections, and the supplement form is trimmed to what a supplement needs with a searchable nutrient picker over a shared catalog of FDA-label micronutrients.

No new module, no new table, no new nav item.

Linked Issue: Closes #1769

## How to Test

1. Check out this branch and run the stack as usual (the migration applies on server boot).
2. Medications page then **+ Add supplement**. Give it a name and a dose form, add a few nutrients (try the multivitamin panel), enter per-dose amounts, save.
3. Verify it appears under **Supplements** and not **Meds**, that the Log view shows it under "Today's supplements", and that the adherence overview follows the filter.
4. Add a schedule, then log a dose from the Log view.
5. Verify Reports > Nutrients shows the logged dose's nutrients, split food vs supplement per nutrient.
6. Edit the supplement's nutrients, then re-check a previously logged day - the historical totals must not move.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

### Before

2026-08-03 18_06_40-supplements-vitamins and 1 more tab - File Explorer

### After

2026-08-03 17_38_50-Greenshot

2026-08-03 17_39_13-Greenshot

2026-08-03 17_39_55-Greenshot

2026-08-03 17_40_07-Greenshot

2026-08-03 17_40_42-Greenshot

</details>

## Notes for Reviewers

**Scope: this is phase one of two, and the split is deliberate.** The rollup mechanism is general — it sums any fixed nutrient field plus custom nutrients — but the *input* surface here is limited to micronutrients, the vitamins and minerals printed on a Supplement Facts label. Nothing in this PR's UI can populate a supplement's energy or macro values, so daily calories are untouched in practice. Exposing those is the follow-up.

**Which surfaces supplements reach.** Supplement contributions appear in `GET /foods/nutrition/today`, the chatbot nutrition tools, and Reports > Nutrients. They deliberately do **not** reach `/api/daily-summary`, which computes the Diary's eaten/net/remaining from food entries alone. Wiring that path belongs with the phase-two change that makes calories enterable — it is the core calorie calculation and shouldn't be touched by a PR that cannot produce a calorie.

**Schema.** Columns only, on two existing tables. No `CREATE TABLE`, so `rls_policies.sql` needs no new policy and both tables are already RLS-covered. Two migrations: the columns plus an `sf_try_numeric()` helper, and a small idempotent seed adding softgel/gummy/powder to `medication_types` (the dose-form select is an FK to it).

**Permissions.** A supplement's nutrient payload is nutrition data, so writing it requires `diary` in addition to the `medications` permission guarding the route. A caregiver with medication access but not diary access sees an explanation in place of the nutrient editor, and the API enforces the same rule rather than trusting the client. It strips the payload instead of rejecting, so the caregiver can still manage the medication: on create the supplement saves without nutrients but keeps its classification, and on update the sparse patch preserves whatever the owner set. Documented in `docs/content/2.features/9.family-friends-sharing.md`.

**Nutrient definitions.** Picking a catalog nutrient find-or-creates the user's own custom nutrient, seeded with the catalog's unit, Daily Value and aliases, so %DV works immediately and provider imports can match it. Creation is deferred to save, so cancelling leaves nothing behind. Deduplication is by name and alias, and nutrients already first-class `food_variants` columns are never shadowed — that would double-count in the report's UNION.

One deliberate trade-off: supplement-seeded nutrients register for the goal and report views but not food search, to avoid cluttering food cards with vitamins tracked only for a supplement. They can be enabled for food search in Settings.

**The second commit is separable.** It reworks the Log view so medications and supplements read as separate cards, makes the log calendar follow the filter, and flows the view as one grid instead of two. It touches no server code. The grid change is a genuine fix — the calendar is taller than the adherence stack, so two stacked grids left a dead gap under the checklist — but it does affect the page for users with no supplements, so it is easy to drop if you would rather it came separately.

Two implementation notes for that commit. The dose card body is a render function rather than a child component, because it closes over the dose handlers and per-row note state and a component would remount the note inputs mid-typing. And entry visibility shares one rule with the Log view's lists, including the part where the mixed view must not filter at all, so an entry that outlived its medication keeps showing up.

**Supplement dose is pinned at 1.** A supplement's nutrient payload is expressed per dose, so pinning keeps "nutrition per dose" unambiguous. This is a scoping choice rather than a hard constraint: the entry snapshot already multiplies by `dose_amount`, so letting supplements carry a real dose works mechanically today and is a natural follow-up once the editor states what the payload is per. The dose fields this dialog now edits are shown for medications and hidden on the supplement form.

Where this branch and `main` had overlapping fixes, `main` wins: the shared `formatDose` helper replaced an equivalent local one this branch had grown, and the Cabinet's dose display now goes through it rather than a supplement-specific special case.

**Nutrient definitions outlive the supplement that introduced them, by design.** Picking a nutrient on a supplement form find-or-creates a `user_custom_nutrients` row for that user and registers it for the goal and report views. Deleting the supplement, or its logged doses, does not remove that definition — so the nutrient keeps appearing in Reports with zero values until it is removed under Settings > Custom Nutrients.

That is deliberate: the definition is user-level and a food or another supplement may reference the same nutrient, so deleting one supplement should not destroy it. The rough edge is the asymmetry — creation is implicit (a side effect of building a supplement) while removal is explicit and lives in Settings. Offering to clean up now-unreferenced nutrients when a supplement is deleted would close that gap; noting it as a follow-up rather than expanding this PR.

**Two things I'd like your call on rather than guessing:**

1. `db_schema_backup.sql` is not regenerated. Running the backup script in a dev environment emits environment-specific application-role GRANT names and unrelated `pg_dump` ordering churn that swamps this PR's schema delta. Happy to produce it however you prefer — a clean throwaway database, or a documented invocation that yields a portable snapshot from a feature branch.

2. Renaming a custom nutrient does not migrate the stored JSONB keys — but that is already true on `main` for `food_variants`, `food_entries`, `goal_presets` and `user_goals`; only the goal-direction preference key is migrated today. I've left the supplement stores consistent with that rather than fixing half of it here, and would rather open a separate PR covering every store at once.

**Known limitation.** `getDailyNutritionTotalsRange` groups from `food_entries`, so a date with logged supplements and no food produces no row. That grouping predates this PR. Surfacing such a day naively would emit a row reading e.g. `calories: 15`, making trend charts dive to near-zero on days the user simply didn't log food — worse than the current gap. Doing it properly needs consumers to distinguish a supplement-only day, which is a payload shape change.

**Tests.** Frontend 701 passing across 80 suites with `validate` clean; server 2680 passing (the 216 skipped are the DB-gated RLS matrix), `validate` clean.

## Future changes

- Phase two: energy and macros in the nutrient picker, plus wiring the Diary summary path.
- Offer to remove now-unreferenced nutrient definitions when a supplement is deleted.
- Barcode import from the NIH DSLD label database (free, no API key, ~200k US labels), with OCR fallback.

## Summary by CodeRabbit

* **New Features**
  * Added supplement management alongside medications, including dose forms, nutrient details, filtering, calendars, and Today’s intake.
  * Added a searchable nutrient picker with aliases, compound nutrients, multivitamin presets, duplicate prevention, and custom nutrient creation.
  * Nutrition reports now separate overall, food, and supplement contributions.
* **Improvements**
  * Supplement nutrients are included in daily totals and preserved with logged entries.
  * Added removable nutrient fields and clearer validation for nutrient and dose values.
  * Improved medication permissions and updated related documentation.


jsandai:
Everything is addressed in `6dd030b` apart from two findings I want to push back on, with reasoning below.

## Fixed

**Diary permission on supplement nutrition writes.** This was the important one and the gap was wider than the middleware I had written. Guarding medication create and update left the logging path open: `createEntry` snapshots the nutrient payload server side, so a caregiver holding medication access but not diary access could log doses of a supplement the owner had already set up, and the report would scale them. Entry create, update and delete, plus schedule creation, now resolve the medication's subtype and reject supplement writes without diary access.

Rejecting rather than stripping here, unlike the medication paths. There is no harmless subset of "log this dose" to let through. Plain medications are unaffected, so a caregiver keeps full access to everything that is not a supplement. The 403 is now in the OpenAPI blocks for those routes.

**Days that have only supplement entries.** Fixed in both queries. They now drive from the union of food entry and taken supplement dates, and left join the food aggregates onto that set, so a day with supplements and no food returns a row with zeroed food totals instead of disappearing.

I had ultimately decided against this previously on the grounds that it makes trend charts dip toward zero on such days. That was the wrong call - the dip is accurate: if no food was logged then the logged intake really is near zero, and a gap hides the fact that anything was logged at all. Note this required qualifying the food columns as `fe.`, since the left join makes bare column names ambiguous.

**Concurrency in catalog provisioning.** Now `INSERT ... ON CONFLICT (user_id, name) DO UPDATE ... RETURNING`, so two simultaneous requests for the same nutrient both receive the row rather than one failing the unique constraint.

**Colliding resolved nutrient values.** Good catch, this silently dropped an amount the user had typed. Two provisional keys can resolve onto one existing nutrient, for example a user's "Vitamin D" absorbing both "Vitamin D2" and "Vitamin D3", and `Object.fromEntries` let the second overwrite the first. The payload is accumulated now.

**`sf_try_numeric` overflow.** Exponent and length are bounded so the cast cannot raise. The helper exists precisely so that a malformed stored value cannot error an entire report, so a hole in it defeated the purpose.

**Explicit `any`, the toast, and the tests.** Removed the suppression in the display preference lookup in favour of a declared row type, replaced the untyped mock clients in the repository tests with a shared typed helper, and routed the catalog error message through `t()`. Removing `any` from the test doubles surfaced two genuine "possibly undefined" results from `.find()`, which are asserted now rather than reached through.

The dose gate tests also assert that `canAccessUserData` is called with `diary` specifically, not just that the request is refused. Asserting the status code alone would still pass if the gate were changed to consult `medications`, which every caller of those routes already holds, so the test would stay green while the hole reopened.

## Not changing

**The branded ID and `user_id` schemas in `Medications.zod.ts` and `MedicationEntries.zod.ts`.** The analysis is correct. `z.string().and(z.object(...))` cannot parse a UUID, and `z.any()` infers as `any`.

However those files carry a `// Generated by ts-to-zod` header, 41 of the 75 files in `shared/src/schemas/database/` use the identical branded intersection pattern, and `userIdSchema = z.any()` is what the generator emits throughout. Both schemas are also type only in practice: `medicationsSchema` and `medicationEntriesSchema` are never parsed at runtime anywhere in the repo.

Hand editing two generated files would make them inconsistent with 41 siblings and would be overwritten the next time anyone regenerates. If these schemas should carry real runtime types then that is a worthwhile change to the generator and all 75 outputs together, and I am happy to open it separately, but it seems wrong to apply it to two files inside a feature PR.

## One thing worth flagging that was not raised

Nutrient definitions outlive the supplement that introduced them. Picking a nutrient find-or-creates a `user_custom_nutrients` row and registers it for the goal and report views, and deleting the supplement or its logged doses does not remove that definition, so the nutrient keeps appearing in Reports with zero values until it is removed under Settings.

That part is deliberate, since the definition is user level and a food or another supplement may reference the same nutrient. The rough edge is the asymmetry: creation is implicit, as a side effect of building a supplement, while removal is explicit and lives in Settings. Offering to clean up nutrients that are no longer referenced when a supplement is deleted would close that gap. It is listed under Future changes rather than expanded here.

## Verification

Frontend: 701 tests across 80 suites, `pnpm run validate` clean.
Server: 2685 passing, 216 skipped (the DB gated RLS matrix), `validate` clean.

@coderabbitai review

jsandai:
Fixed in `e3cae83`.

`PUT /v2/medications/schedules/:id` now carries the same supplement diary check, resolving the subtype through the schedule ID. The chain is exactly as you describe: `updateSchedule` writes `dose_amount`, `createEntry` reads `COALESCE(ms.dose_amount, m.dose_amount)` and stores it as `dose_amount_snapshot`, and the report multiplies the nutrient payload by it. A test asserts the update returns 403 without diary access, that `updateSchedule` is not called, and that the check asks for `diary` specifically.

While adding it I swept the rest of the router and found one more of the same kind that had not been raised: `DELETE /v2/medications/schedules/:id`. Deleting a schedule drops its dose override, so future entries fall back to the medication dose, which moves the multiplier just as editing it does. That route now carries the gate as well.

Six routes are gated now: entry create, update and delete, schedule create, update and delete. Medication create and update keep the strip behaviour rather than a rejection, since there a caregiver has legitimate reason to rename or reschedule and only the nutrient payload needs withholding.

Frontend: 701 tests across 80 suites, `validate` clean.
Server: 2686 passing, 216 skipped, `validate` clean.

@coderabbitai review

jsandai:
Fixed in `8dc6091`.

The guard ran ahead of each route's `UuidParamSchema` and its lookups compare against uuid columns, so a malformed id reached Postgres and raised where the route previously returned a clean 400. The guard now shape-checks the id first and passes anything invalid straight through to that existing validation, so the 400 comes from the same place it always did.

Applied to every id the guard reads, not just the schedule routes: the entry id on update and delete, and the medication id in the entry-create body were exposed the same way.

Three tests cover it, asserting a 400 for a malformed id on schedule update, schedule delete and entry update, and asserting that `canAccessUserData` is never reached. That second assertion is the one that actually pins the ordering, since it fails if the guard starts doing work before the id is known to be well formed.

Frontend: 701 tests across 80 suites, `validate` clean.
Server: 2689 passing, 216 skipped, `validate` clean.

@coderabbitai review

CodeWithCJ (commented):
Could you merge the migration scripts into a single SQL file

SparkyFitnessFrontend/src/pages/Medications/NutrientPicker.tsx
```diff
@@ -0,0 +1,445 @@
...
+        </PopoverContent>
+      </Popover>
+
+      <Button
+        type="button"
+        variant="ghost"
+        size="sm"
+        onClick={addMultivitaminPanel}
+      >
+        <Sparkles className="mr-2 h-4 w-4" />
```
apedley: Sparkles usually means AI. Is there a better icon?

jsandai:
> Could you merge the migration scripts into a single SQL file

Done

SparkyFitnessServer/routes/v2/medicationRoutes.ts
```diff
@@ -547,7 +713,11 @@ router.delete(
 );
 
 router.get('/:id', getMedication);
-router.put('/:id', updateMedication);
+router.put(
```
apedley, line 735: Also need to strip dose_amount here when the target medication is a supplement

SparkyFitnessServer/routes/v2/medicationRoutes.ts
```diff
@@ -39,6 +41,148 @@ const router = express.Router();
...
+          WHERE ms.id = $1 AND ms.user_id = $2`,
+        [scheduleId, userId]
+      );
+      return Boolean(res.rows[0]?.is_supplement);
+    }
+    if (entryId) {
+      const res = await client.query(
+        `SELECT m.is_supplement
+           FROM medication_entries me
+           JOIN medications m ON m.id = me.medication_id
```
apedley: medication_entries.medication_id is ON DELETE SET NULL, and entries from a deleted supplement keep feeding the report through nutrients_snapshot (the queries only check status + snapshot). This INNER JOIN returns false for those orphans, so a meds-only caregiver can update/delete/status-flip that history - a nutrition write without diary access.

LEFT JOIN and treat `m.is_supplement OR me.nutrients_snapshot IS NOT NULL` as supplement. The snapshot is what actually feeds the report so it's arguably the more correct test anyway.

SparkyFitnessServer/services/customNutrientService.ts
```diff
@@ -44,14 +61,27 @@ class CustomNutrientService {
...
     const client = await getClient(userId);
     try {
       const id = uuidv4();
       const result = await client.query(
+        // ON CONFLICT keeps this idempotent under concurrency. Catalog provisioning
+        // does a read-before-create, so two simultaneous requests for the same nutrient
+        // can both see nothing and race to insert; without this the unique
+        // (user_id, name) constraint fails one of them instead of returning the row
+        // that now exists. The no-op DO UPDATE is what makes RETURNING yield it.
         `INSERT INTO user_custom_nutrients (id, user_id, name, unit, aliases)
```
apedley, line 87: This fixes the race but opens a non-race regression. Before, inserting a duplicate name raised the unique violation and stopped. Now it returns the existing row and falls through to the goal seeding, so the jsonb_set calls overwrite the user's tuned target for that nutrient in goal_presets and future user_goals (with 0 from the settings page path, the DV from the catalog path), and re-add it to view groups they may have deliberately hidden it from.

The race you're guarding writes identical values from both sides, so distinguishing is enough:

```sql
RETURNING *, (xmax = 0) AS inserted
```

and skip the view-group + goal seeding when it wasn't inserted.

SparkyFitnessFrontend/src/pages/Reports/NutritionChartsGrid.tsx
apedley, line 92: T he domain is computed from the total series only, and the two new split lines are always well below the total. Calcium with food 800 and supplement 300 totals 1100, which gives a domain around [770, 1130] - the supplement line at 300 never draws inside the plot.

SparkyFitnessFrontend/src/pages/Reports/NutritionChartsGrid.tsx
```diff
@@ -192,10 +191,18 @@ const NutritionChartsGrid = ({
                     <CardTitle className="text-sm">
                       {chart.label} ({chart.unit})
                     </CardTitle>
-                    <span className="text-xs text-muted-foreground font-normal">
-                      {t('reports.average', 'Avg')}: {formattedAverage}{' '}
-                      {chart.unit}
-                    </span>
+                    <div className="text-right text-xs text-muted-foreground font-normal">
+                      <div>
+                        {t('reports.average', 'Avg')}: {formattedAverage}{' '}
```
apedley, line 205: Two things about this line. It renders for everyone - the server always emits supplement_* = 0, so every existing user's every chart grows a "Supplements: 0" header line and extra tooltip rows.

And I think the averages are the wrong measurement. People care about totals and % not average.

apedley (changes requested):
To answer your questions:

1. For the db schema backup I've just been editing it by hand. Someday soon I'll write the script to auto generate it in CI. 
2. renaming custom nutrients not migrating stored JSONB keys is a pre existing problem across all the stores and a separate PR

On the ts-to-zod pushback, those headers are stale. Nothing regenerates the files in schemas/database they are hand maintained.

Left a few comments here are smaller findings:

1. AddMedicationDialog.tsx:475 force-rewrites dose_amount to 1 on every supplement edit, not just create. Once dose is gated server-side, preserving
editMed.dose_amount on edit is the safer shape.
2. The controlled open/onOpenChange mode added to AddMedicationDialog has no consumers - YAGNI unless something is about to use it.
3. Saving a supplement with N new free-text nutrients pops N "Custom nutrient added" toasts before the save toast.
4. MedicationNutrients in types/medications.ts hand-lists the 17 keys a third time - could be `Partial<Record<FoodVariantNutrientField, number>> & {
custom_nutrients?: Record<string, number> }`.
5. nutrientPickerUtils.ts is two functions with one consumer - medicationUtils.ts already hosts the other helpers from this PR.

CodeWithCJ:
Don't forget to resolve the conversations to merge this PR

jsandai:
@apedley thanks for your review. Taking the two answers first: I'll edit `db_schema_backup.sql` in this PR, and I'll leave the JSONB rename problem alone as a separate change.

On the ts-to-zod pushback, you are right. I read the header and reasoned from it without checking whether anything actually runs the generator.

## The charts

You're right that averaging the split is the wrong measurement, and chasing that led somewhere better than swapping in totals.

The question behind a food versus supplement split is what proportion comes from a pill, so the header now shows a single share rather than two more averages. A mean is also actively misleading for the supplement arm on intermittent dosing: take something three days in seven and the average reads a number that describes neither the days you took it nor the days you did not. A share does not have that problem, and it stays comparable when the date range changes, which a total would not.

I kept the existing per day average for the nutrient itself. Nutrition targets are daily, so that number sits next to the goal and means something, whereas a total scales with the window and invites comparing a 30 day figure against a 7 day one.

Two of your other findings fall out of this rather than needing separate fixes:

The split series are gone, so the y domain problem goes with them. You were right that lines well below a domain computed from the total never draw, and once the split is a header stat there is nothing to scale.

The share is hidden entirely when nothing was supplemented. Existing users who track no supplements now see their charts exactly as before, with no zero line and no extra tooltip rows.

## The icon

Agreed, and I had thought about that too. Swapped for `ListPlus` - I considered a pie chart and decided against it: it signals viewing a breakdown, which is what the report does, so on a button that mutates the form it points the wrong way. `ListPlus` says what the button actually does, which is add a set of rows in one click, and it sits naturally beside the `Plus` on the single nutrient action.

Frontend: 701 tests across 80 suites, `validate` clean.

2026-08-04 17_48_42-Settings

apedley (approved):

PR #2036, opened 2026-08-04
fix(medications): solve the absorption rate so the modeled curve peaks at tMax

### LINKED ISSUE #2035: Modeled serum curve does not peak at the configured time-to-peak (absorption rate is not solved for tMax)
### What happened?

The modeled serum-level curve does not peak at the `tMaxDays` value configured for the drug. The absorption rate constant is derived in a way that does not solve for the peak, so every profile's modeled peak lands substantially later than the time-to-peak it was given.

The curve shape is therefore wrong in a way the parameters say it should not be — the field is documented as `time to peak concentration in days`, and the code comment states the rate is "derived so the single-dose peak lands near tMax".

### Cause

In `shared/src/medications/glp1.ts`:

```ts
const ke = eliminationRate(profile.halfLifeDays);
// Absorption rate: derived so the single-dose peak lands near tMax.
const ka =
  profile.tMaxDays > 0
    ? Math.max(ke * 1.5, Math.LN2 / profile.tMaxDays)
    : ke * 4;
```

`Math.LN2 / tMax` is a half-life-style conversion, not the peak of a Bateman curve.

For a one-compartment model with first-order absorption, the peak occurs where the absorption rate equals the elimination rate:

```
ka·e^(−ka·t) = ke·e^(−ke·t)
```

Solving for `t` gives the standard result

```
tmax = ln(ka / ke) / (ka − ke)
```

which has no closed form in `ka` and so has to be solved numerically. Substituting `ka = ln2/tMax` into that expression does not return `tMax`.

Reference for the relation: it is the standard extravascular one-compartment result, obtained by setting `dC/dt = 0` on the Bateman function `C = (F·D·ka)/(V·(ka−ke))·(e^(−ke·t) − e^(−ka·t))`. See e.g. [Extravascular Administration — Pharmacokinetics, University of Lausanne](https://sepia2.unil.ch/pharmacology/profiles/extravascular-administration/) and [Tmax — REVIVE/GARDP](https://revive.gardp.org/resource/tmax-time-to-peak-drug-concentration/?cf=encyclopaedia). Background on the one-compartment first-order-absorption model itself: Weiss M., *Is the One-Compartment Model with First Order Absorption a Useful Approximation?*, Pharmaceutical Research 40(9):2147–2153, 2023, [doi:10.1007/s11095-023-03582-1](https://doi.org/10.1007/s11095-023-03582-1).

### Actual vs expected

Computed from the shipped profiles by sampling `serumLevelAt` for a single dose:

| `halfLifeDays` | configured `tMaxDays` | modeled peak |
|---|---|---|
| 7 | 1.5 | **4.24 d** |
| 5 | 1.5 | **3.72 d** |
| 4.7 | 2.0 | **4.29 d** |
| 0.54 | 0.46 | 0.63 d |

Expected: the modeled peak matches the configured `tMaxDays` for each profile.

### Steps to reproduce

```ts
import { serumLevelAt } from '@workspace/shared';

const profile = { halfLifeDays: 7, tMaxDays: 1.5 };
let best = { day: 0, level: -Infinity };
for (let day = 0; day <= 40; day += 0.01) {
  const level = serumLevelAt(day, [{ day: 0, doseMg: 1 }], profile);
  if (level > best.level) best = { day, level };
}
console.log(best.day); // 4.24, expected ~1.5
```

### Effect in the UI

Smaller than the single-dose numbers suggest, which is probably why it has gone unnoticed. Under repeat dosing each new dose rises against the decaying tail of the previous ones, so the visible local peaks arrive earlier than an isolated dose would: for a 7-day half-life dosed weekly, the local peak converges to ~2.5 days after each dose rather than 4.24. The accumulation envelope and the approach to steady state are unaffected — those follow from the elimination term alone.

So this mainly affects the early part of the curve, single-dose or newly-started schedules, and any profile with a short half-life relative to its cadence.

### Version

v1.6.1

### Anything else?

Happy to open a PR — I have it working locally with tests, solving `ka` numerically so the peak lands on the configured `tMaxDays`. The drug parameter values themselves are untouched; this only makes the model honor the values already there.

JonZavialov:
## Description

The modeled serum curve didn't peak at the `tMaxDays` configured for the drug. `ka` was derived as `Math.LN2 / tMaxDays`, which is a half-life-style conversion rather than a solve for the peak, so every profile peaked later than the value it was given — 4.24 d against a configured 1.5 d for a 7-day half-life.

For a one-compartment model with first-order absorption the peak sits at `ln(ka/ke)/(ka − ke)`, which has no closed form in `ka`. Added `absorptionRate(halfLifeDays, tMaxDays)` that solves it by bisection and used it in `serumLevelAt`.

The expression decreases monotonically in `ka` and tends to `1/ke` as `ka → ke`, so a requested peak later than `1/ke` is unreachable for any valid `ka` and falls back to the slowest absorption the model supports. Results are memoized, since `serumLevelAt` is called once per sample point.

**No drug parameter values are changed** — this only makes the model honor the `tMaxDays` already configured.

Linked Issue: Closes #2035

## How to Test

1. Check out this branch and run `pnpm test` in `SparkyFitnessServer/`.
2. Or sample the curve directly:

```ts
import { serumLevelAt } from '@workspace/shared';

const profile = { halfLifeDays: 7, tMaxDays: 1.5 };
let best = { day: 0, level: -Infinity };
for (let day = 0; day <= 40; day += 0.01) {
  const level = serumLevelAt(day, [{ day: 0, doseMg: 1 }], profile);
  if (level > best.level) best = { day, level };
}
console.log(best.day); // 1.50 on this branch, 4.24 on main
```

3. In the UI, open a medication's modeled curve and confirm the first dose rises to its peak at the configured time-to-peak rather than several days late.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

Pure model change — no UI files touched. The visible effect is described under "Scope of the change" below.

</details>

## Summary by CodeRabbit

* **Improvements**
  * Improved GLP-1 medication serum-level calculations to more accurately reflect configured peak timing.
  * Enhanced consistency across supported medication profiles.
  * Added reliable fallback handling for missing, invalid, zero, unreachable, or extremely small peak-time settings.
  * Improved support for higher-dose medications while maintaining stable, finite serum-level results.

### DIFF
--- a/SparkyFitnessServer/tests/glp1Logic.test.ts
+++ b/SparkyFitnessServer/tests/glp1Logic.test.ts
@@ -2,13 +2,34 @@ import { describe, expect, it } from 'vitest';
 import {
   resolveGlp1Profile,
   eliminationRate,
+  absorptionRate,
   serumLevelAt,
   simulateSerumCurve,
   suggestNextSite,
+  GLP1_DRUG_PROFILES,
   INJECTION_SITES,
   SITE_REST_DAYS,
   type DoseEvent,
 } from '@workspace/shared';
+
+function singleDosePeakDay(
+  halfLifeDays: number,
+  tMaxDays: number,
+  step = 0.005
+): number {
+  const doses: DoseEvent[] = [{ day: 0, doseMg: 1 }];
+  const profile = { halfLifeDays, tMaxDays };
+  let bestDay = 0;
+  let bestLevel = -Infinity;
+  for (let day = 0; day <= halfLifeDays * 6; day += step) {
+    const level = serumLevelAt(day, doses, profile);
+    if (level > bestLevel) {
+      bestLevel = level;
+      bestDay = day;
+    }
+  }
+  return bestDay;
+}
 
 describe('GLP-1 PK helpers', () => {
   it('resolves a profile by id and by brand name', () => {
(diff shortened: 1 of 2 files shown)


CodeWithCJ:
Could you take a look at Coderabbit comment . All comments must be in resolved state for us to merge it.

apedley (commented):
Thanks for fixing this. Everything looks right but I'd drop the cache (shared/src/medications/glp1.ts line 98). It has user typed keys so it keeps growing. it's also module level global mutable state and the file is supposed to be "Pure functions + published reference data — no DB, no side effects." as it says at the top. The speed we get from the cache would be negligible.

apedley (approved):

PR #2120, opened 2026-08-14
feat(medications): supplement energy and macros in the diary

### LINKED ISSUE #2097: [Feature]: Supplement energy and macros in the diary (phase two of supplement tracking)

Yes, I'd like to submit a PR!

Supplement tracking shipped in #2031 with micronutrient input only, and the PR framed that as phase one of two. A supplement can carry vitamins and minerals, but there is nowhere to record the energy or macros printed on the same label.

Plenty of supplements carry real amounts. Fish oil is around 15 kcal per capsule, and a serving of collagen or a greens powder can be 30 to 60 kcal with several grams of protein or carbohydrate. Someone taking a few of those daily is off by a meaningful amount and has no way to correct it inside the app. Logging the supplement as a food instead gets the calories counted but loses the scheduling, reminders and adherence that made it a supplement in the first place.

Let a supplement's nutrition panel accept energy and macros in addition to micronutrients, and have those reach the diary the way food does.

Implementation:

1. An energy and macros group in the supplement nutrient picker, covering calories, protein, carbohydrates, fat and dietary fiber.
2. Logged supplement doses counted in the diary's eaten calories, remaining calories and macro totals, on web and mobile.
3. The supplement contribution shown as its own line, so the headline total still reconciles against the food rows underneath it rather than looking wrong.

Most of the code for this already exists. `getDailyNutritionSummary` has summed a supplement's calories, protein, carbs, fat and fiber since #2031 merged, and the `nutrients` payload has no key restriction, so no schema change is needed. The gaps are the input surface and the diary, which computes its own totals from food entries independently of that query.

Web frontend, Mobile app (iOS), Mobile app (Android), Server / API, Nutrition tracking

**Log the supplement as a food as well.** Gets the calories counted, but you maintain the same product in two places and lose the adherence view. It also double counts if you forget to remove one.

**Leave supplement energy untracked.** Defensible for a multivitamin at 2 kcal, but not for someone taking several softgels and a scoop of something daily.

**A per supplement "count toward calories" toggle.** I considered this and then dropped it. If you don't want a supplement's calories counted, the existing mechanism is to leave its nutrition panel empty, and a flag would let the stored record claim 15 kcal while the totals claim zero. Honouring it would also mean threading it through every rollup and snapshotting it per entry, since toggling it later must not rewrite logged history.

### Additional context

_No response_

jsandai:
## Description

Supplements shipped with micronutrient input only, so there's nowhere to record the energy and macros printed on the same label, and a few fish oil softgels or a scoop of collagen go uncounted. This is phase two, which #2031 described as the follow-up.

An "Include macros" checkbox on the supplement form reveals calories, protein, carbs, fat and fiber as one block, and the diary now counts logged supplement doses. The rollup needed no work: `getDailyNutritionSummary` has summed those five for supplements since phase one, and the `nutrients` payload has no key restriction, so there is no migration. What was missing was somewhere to enter the values, and the diary, which derives its own totals from food entries independently of that query.

Linked Issue: Closes #2097

## How to Test

1. Check out this branch and run the stack as usual. No migration is needed.
2. Medications page, **+ Add Supplement**. Give it a name and a form, then tick **Include macros** and enter the label's per serving amounts. Add a vitamin or two with **Add nutrient** as well.
3. For a countable form (tablet, capsule, softgel, gummy) the panel offers **1 serving = N**. Set it to whatever the label's serving size is.
4. Save, add a schedule, then log a dose from the Log view.
5. Open the Diary. Eaten calories include the dose, and so do the nutrition summary's protein, carbs, fat and fiber.
6. Undo the dose. The totals drop back immediately rather than after a refresh.
7. On mobile, repeat 4 to 6. The dashboard calorie ring, the macro pills and the nutrition details screen should agree with each other and with web.
8. Set the schedule's **Servings per log** above 1 and reopen it. The form states the multiplication that logging will apply.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

- [x] **[MANDATORY for Backend changes] Database Security**: I have updated `rls_policies.sql` for any new user-specific tables. *(No new tables and no schema change at all, so no policy is needed.)*

## Screenshots

<details>

### Before

2026-08-14 13_32_44-Greenshot

### After

2026-08-14 13_26_07-Nextcloud

2026-08-14 13_27_26-Nextcloud

2026-08-14 15_51_37-Greenshot

</details>

## Notes for Reviewers

**No migration.** `medications.nutrients` and `medication_entries.nutrients_snapshot` carry shape-only CHECK constraints with no key restriction, so energy and macros store on the existing columns. Nothing about an existing supplement changes until someone enters a value.

**Why macros are a checkbox rather than picker rows.** They're a closed set that co-occurs: a Supplement Facts label showing calories almost always prints other macros too. Search is the right affordance for a 33 entry catalog and the wrong one for a fixed five, and listing them there pushed the vitamins, which are the common case, below the fold. `MACRO_PICKER_FIELDS` is likewise separate from `MICRONUTRIENT_CATALOG`, whose entries all carry a Daily Value and are candidates for seeding as `user_custom_nutrients`; these five are existing `food_variants` columns that store directly with no provisioning.

Unticking the box drops the keys rather than only hiding the fields, since `getNutrients` reads selected rows only. A collapsed block that went on counting would be exactly the "record says 15 kcal, totals say zero" split this feature exists to avoid.

**Known limitation: platform health writeback stays food-only.** Both writeback paths build their records from the day's food entries, so supplement nutrition does not reach Apple Health or Health Connect. That's not new here: `HC_NUTRIENT_COLUMNS` already writes sodium, calcium, iron, vitamin C and vitamin A, which supplements have been able to carry since #2031, so writeback has been food-only for as long as supplements have existed. Closing it means synthesising a record whose name, meal slot, time anchor and client record id are all product decisions, and the summary exposes day aggregates rather than per-dose entries. Happy to do it, but it seemed better as its own change than improvised inside this one.

**Scoped to five fields on purpose.** A supplement's sodium is not summed by the nutrition summary or the chatbot, so counting it in the diary alone would trade one inconsistency for a subtler one. If the rollup grows, the input should grow with it in the same change.

**Why mobile is in scope.** Its dashboard ring reads the server's `calorieBalance`, which now includes supplements, while the macro pills derive from food entries and the nutrition details screen reads a food-only `caloriesConsumed`. Shipping the server change alone would have left two mobile screens disagreeing about the same day. The normalisation lives in `shared` as `resolveSupplementTotals`, used by both clients, since each deriving its own totals is exactly why they drifted; it also degrades to zeros against a server too old to send them.

Mobile dose mutations now invalidate the daily summary as well. Mobile queries have an infinite stale time, so a missing invalidation is not a brief flicker, it is a ring that never updates. Web already did this.

**Serving, not dose.** A Supplement Facts panel states amounts per serving, and a serving is often more than one tablet/capsule/softgel. Calling the field a dose invites reading it as tablets etc., entering the per serving amounts and then logging 2, which doubles the panel. Supplement surfaces now say serving; the shared medication vocabulary is untouched, because a medication genuinely has doses and the calendar, reports and log views are common to both.

A supplement can also state how many units make up a serving, for the forms where counting means something. It is descriptive only: nutrition is entered and stored per serving either way, so it changes no arithmetic, and it lives in `custom_fields` next to the GLP-1 metadata. `formatDose` normalises the stored placeholder, so a cabinet stocked either side of the rename reads consistently without migrating any rows.

**Known limitation, unchanged by this PR.** A family delegate with diary access but not medication access sees zeros for supplement nutrition, because `medication_entries` is gated by `has_medication_read_access` and the aggregate runs under the actor's RLS context. That's pre-existing: the nutrition summary and the Reports query read the same table the same way, so such a delegate already sees no supplement nutrition anywhere. Nothing that worked for them stops working here.

I left that alone deliberately. Supplement nutrition derives from medication rows the owner chose not to share, so surfacing it would tell that delegate the owner takes supplements and how much energy they carry. Changing it needs a `SECURITY DEFINER` path returning aggregate nutrition without exposing medication records, which is a decision about what the diary permission means and would apply to all three read sites.

**Supplement authoring on mobile.** The mobile medication form has no supplement fields, so a supplement is created and its nutrition edited on web, while mobile logs doses of it and, with this PR, counts them. That split predates this change: `MedicationFormScreen` has never carried `is_supplement` or `nutrients`, so it reads the same on `main` today. Happy to bring the fields across as a follow-up for parity. Editing a medication from mobile is safe for supplements either way, since the update writes only the keys the payload actually contains.

## Summary by CodeRabbit

* **New Features**
  * Supplements now support calories, macros, serving units, and scalable dose tracking.
  * Supplement nutrition is included in daily calorie, macro, fiber, dashboard, and diary totals.
  * Supplement-only diary days now display nutrition summaries instead of empty states.
  * Added clearer serving labels, pluralization, and nutrition guidance.
* **Bug Fixes**
  * Improved refresh behavior after medication entries are created, updated, deleted, or logged from notifications.
  * Prevented duplicate macro and micronutrient selections.


jsandai:
Two of the three are real and are fixed; the third does not reproduce, and I've taken the suggested change anyway as future-proofing. Detail below.

**Reject zero units per serving** (`AddMedicationDialog.tsx`) - fixed.

Confirmed: the input carried `min="0"` while save keeps the value only when `units > 0`, so a typed zero was accepted on screen and then silently dropped. Changed to `min="1"`.

**Use "serving" in the scaling message** (`ScheduleManager.tsx`) - fixed, though the diagnosis is slightly different from what was reported.

The rendered English already said "servings", because `translation.json` carries the corrected values. What still said "doses" was the inline fallback default in the source, which had drifted out of sync with the resource. Anyone hitting the fallback path saw the wrong noun for a value that is per serving, which is exactly the capsule-count confusion described. Both fallbacks now match the resource: "Each log counts {{dose}} servings…" and "…the nutrition entered per serving."

**Filter macro short labels from custom nutrient options** (`NutrientPicker.tsx`) - does not reproduce, but change taken.

`normalizeNutrientName` lowercases, and every macro's `shortLabel` already normalizes onto that macro's own `displayName` or one of its aliases, so all five are in `catalogNames` as it was built before this change:

```
"Calories" -> "calories"  (= displayName)
"Protein"  -> "protein"   (= displayName)
"Carbs"    -> "carbs"     (in carbs aliases)
"Fat"      -> "fat"       (= displayName)
"Fiber"    -> "fiber"     (in dietary_fiber aliases)
```

The cited case is one of these: a custom nutrient named "Carbs" normalizes to `carbs`, which is already in the set via `aliases: ["carbs", "total carbohydrate", "carbohydrate"]`, so it was already filtered out of the picker and could not be saved into `custom_nutrients` alongside the real carbs field.

I've added `field.shortLabel` regardless. It is a no-op today, but it keeps this set aligned with `resolveMacroFieldKey`, which matches on the same four properties, so a macro added later whose shortLabel is not also a displayName or alias cannot slip through and be counted separately. The reasoning is recorded in a comment at the call site so the next reader does not mistake it for a live fix.

Flagging the non-reproduction explicitly rather than quietly accepting the diff, since the merge-risk note is gated on that finding and I would rather the record be accurate.

All three changes are in a new commit on top (`8bac9a4`), not an amend. Full suite green at that head: frontend 857 across 96 suites, server 2987 passing with 216 skipped, mobile 4985 across 298 suites, all three `validate` clean.

@coderabbitai review

CodeWithCJ (approved):

PR #2123, opened 2026-08-15
fix(ai): support current-generation Claude models on the Anthropic provider

### LINKED ISSUE #2077: [Bug]: Anthropic provider can't use Claude Opus 5 / Sonnet 5 — hardcoded max_tokens 2048 with no thinking parameter truncates the response
### Summary

The Anthropic provider can't be pointed at Claude Opus 5 or Sonnet 5. Two hardcoded values in the request builder combine so that any current-generation model would truncate its response, and the model picker doesn't offer them anyway.

Version: v1.6.1 (NixOS module deployment, native systemd, Anthropic-only provider).

### The blocker

`buildAnthropicRequest` sends `model`, `max_tokens`, `messages`, and — for structured extraction — a forced tool call, but never a `thinking` parameter:

- `SparkyFitnessServer/ai/providerDispatch.ts:98` — `const ANTHROPIC_MAX_TOKENS = 2048;`
- `SparkyFitnessServer/ai/providerDispatch.ts:548` — `max_tokens: ANTHROPIC_MAX_TOKENS`
- `SparkyFitnessServer/ai/providerDispatch.ts:556-564` — `tools: [...]` + `tool_choice: { type: 'tool', name: ctx.toolName }`

On Claude Opus 4.8 and earlier, omitting `thinking` means thinking is off, so all 2048 tokens go to the answer. On **Claude Opus 5 and Sonnet 5, omitting `thinking` runs adaptive thinking by default**, and `max_tokens` caps thinking *and* the visible response together. At 2048 tokens with a forced tool call, reasoning consumes the budget and the tool response gets cut off — surfacing through the existing `stop_reason === 'max_tokens'` path at `providerDispatch.ts:706-710` as "AI service truncated the response".

So switching the model string alone isn't enough: the newer models need either a raised `max_tokens` or an explicit `thinking: { type: 'disabled' }`.

### Secondary: the model list is a generation behind

`SparkyFitnessFrontend/src/utils/aiServiceUtils.ts:53`

```ts
case 'anthropic':
  return ['claude-sonnet-4-6', 'claude-haiku-4-5', 'claude-opus-4-8'];
```

This matters more than it looks for the label-scan and food-photo features, because it's a **vision** constraint rather than a reasoning one. High-resolution image support (2576px on the long edge, vs 1568px) landed with Opus 4.7 and, in the Sonnet line, only with Sonnet 5. Of the three options above, `claude-opus-4-8` is the only one that has it — so anyone who wants high-res label OCR is currently forced onto the most expensive of the three, and `claude-sonnet-5` (same capability, Sonnet pricing) isn't selectable.

I confirmed the server does no downscaling before dispatch — the only image processing dependency is `heic-convert` for HEIC→JPEG — so the captured resolution does reach the model and the model's own ceiling is what binds.

### Suggested fix

1. Send `thinking: { type: 'disabled' }` on the Anthropic structured-extraction path (label scan, photo estimation, unit conversion), where reasoning tokens aren't wanted and the output is a constrained tool call. Alternatively raise `ANTHROPIC_MAX_TOKENS` and leave thinking adaptive.
2. Make `ANTHROPIC_MAX_TOKENS` configurable rather than a module constant — 2048 is tight for a structured nutrition payload even without thinking.
3. Add `claude-opus-5` and `claude-sonnet-5` to `getModelOptions('anthropic')` once (1) is in.

### One thing to watch when adding Opus 5

`providerDispatch.ts:551-553` only sends `temperature` when `ctx.temperature !== undefined`, and nothing in the current call paths sets it — so this is fine today. But **Claude Opus 5 rejects `temperature` outright with a 400** (Sonnet 5 rejects non-default values). If a temperature setting is ever surfaced in the AI provider UI, it'll need to be suppressed for those models.

vaibhav8a:
## Description

The Anthropic provider could not be pointed at current Claude models: `max_tokens` was hardcoded to 2048 (truncating responses), `temperature` is rejected outright by Opus 4.7+, and Opus 5 / Sonnet 5 were missing from the model list.

**1. `max_tokens` 2048 → 8192**, overridable via `SPARKY_FITNESS_ANTHROPIC_MAX_TOKENS`. On Opus 5 / Sonnet 5, omitting `thinking` runs adaptive thinking by default and `max_tokens` caps thinking *and* the visible response together, so a forced tool call at 2048 could have its reasoning eat the budget and truncate — surfacing via the existing `stop_reason === 'max_tokens'` handler as "AI service truncated the response".

I deliberately did **not** take the issue's first suggestion (`thinking: {type: 'disabled'}` on the structured-extraction path). Disabling thinking there has a documented failure mode: the model can write the tool call into its visible text instead of emitting a `tool_use` block, so the turn completes with no error and the call silently never runs — exactly the forced-tool-call path this provider depends on.

**2. `temperature` dropped for models that reject it.** The issue calls this a future concern ("nothing in the current call paths sets it"), but that isn't right: `foodOptionsService` sends `temperature: 0.7` on the Anthropic path, with a test asserting it. Since `claude-opus-4-8` was already in the model list, that path fails today for anyone who picked it.

**3. Model list** gains `claude-opus-5` and `claude-sonnet-5`. This matters for label-scan and food-photo, not just reasoning: high-resolution vision (2576px long edge vs 1568px) is on Opus 4.7+ and, in the Sonnet line, only from Sonnet 5 — so `claude-opus-4-8` was previously the only high-res option, forcing the most expensive model for label OCR.

Linked Issue: Closes #2077

## How to Test

1. Check out this branch, then in `SparkyFitnessServer/` run `pnpm run typecheck && pnpm run lint && pnpm run test`.
2. The three `max_tokens` cases in `providerDispatch.test.ts` fail on `main` (`expected 2048 to be 8192`) and pass here.
3. With an Anthropic key, select `claude-opus-5` or `claude-sonnet-5` and run a food-description or label-scan request; previously the model was unavailable, and `claude-opus-4-8` returned a 400 from the `temperature` on the food-options path.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

- [x] **[MANDATORY for Frontend changes] Translations**: No translation files touched — the only frontend change is the model list in `src/utils/aiServiceUtils.ts`.

- [x] **[MANDATORY for Backend changes] Code Quality**: I have run typecheck, lint, and tests. No new endpoints or files — this changes request construction in an existing dispatcher.
- [x] **[MANDATORY for Backend changes] Database Security**: No new tables, so `rls_policies.sql` needs no update.

## Test Results

Added to `providerDispatch.test.ts`:

- `max_tokens` defaults to 8192, honours the env override, and falls back on a non-numeric value.
- `temperature` omitted for `claude-opus-5` / `claude-opus-4-8` / `claude-sonnet-5`; still sent for `claude-sonnet-4-6` / `claude-haiku-4-5`.

Also updated the stale assertion in `foodOptionsService.test.ts`, which pinned the old 2048.

```
SparkyFitnessServer:  3002 passed, 216 skipped, 0 failed
                      typecheck + lint clean
SparkyFitnessFrontend: pnpm run validate  (typecheck + lint + format:check) clean
```

No UI changes, so no screenshots apply.

## Summary by CodeRabbit

* **New Features**
  * Added support for Claude Sonnet 5 and Claude Opus 5 in the AI model selection options.
* **Improvements**
  * Increased the maximum length of responses from Anthropic models.
  * Improved compatibility across Claude model versions by applying supported request settings automatically.
* **Bug Fixes**
  * Prevented unsupported configuration options from being sent to Claude models that reject them.

### DIFF
--- a/SparkyFitnessFrontend/src/utils/aiServiceUtils.ts
+++ b/SparkyFitnessFrontend/src/utils/aiServiceUtils.ts
@@ -50,7 +50,13 @@ export const getModelOptions = (serviceType: string): string[] => {
         'gpt-5.4',
       ];
     case 'anthropic':
-      return ['claude-sonnet-4-6', 'claude-haiku-4-5', 'claude-opus-4-8'];
+      return [
+        'claude-sonnet-5',
+        'claude-sonnet-4-6',
+        'claude-haiku-4-5',
+        'claude-opus-5',
+        'claude-opus-4-8',
+      ];
     case 'google':
       return [
         'gemini-2.5-flash',
--- a/SparkyFitnessServer/ai/providerDispatch.ts
+++ b/SparkyFitnessServer/ai/providerDispatch.ts
@@ -95,8 +95,33 @@ export type DispatchResult =
 
 const DEFAULT_TIMEOUT_MS = 90_000;
 const OLLAMA_DEFAULT_TIMEOUT_MS = 120_000;
-const ANTHROPIC_MAX_TOKENS = 2048;
+// On Claude Opus 5 and Sonnet 5, omitting the `thinking` parameter runs
+// adaptive thinking by default, and max_tokens caps thinking *and* the visible
+// response together. At 2048 with a forced tool call, reasoning could consume
+// the budget and truncate the tool response, surfacing as stop_reason
+// 'max_tokens'. 2048 was also tight for a structured nutrition payload even
+// without thinking. Override with SPARKY_FITNESS_ANTHROPIC_MAX_TOKENS.
+const ANTHROPIC_MAX_TOKENS_DEFAULT = 8192;
+const ANTHROPIC_MAX_TOKENS = (() => {
+  const raw = process.env.SPARKY_FITNESS_ANTHROPIC_MAX_TOKENS;
+  if (!raw) return ANTHROPIC_MAX_TOKENS_DEFAULT;
+  const parsed = parseInt(raw, 10);
+  return Number.isFinite(parsed) && parsed > 0
+    ? parsed
+    : ANTHROPIC_MAX_TOKENS_DEFAULT;
+})();
 const ANTHROPIC_VERSION = '2023-06-01';
+
+// Claude Opus 4.7 and later reject `temperature` outright with a 400, and
+// Sonnet 5 rejects non-default values. Sending it to those models fails the
+// request before any work happens, so it is dropped rather than forwarded.
+// Older models (Sonnet 4.6, Haiku 4.5, and earlier) still honor it.
+const ANTHROPIC_MODELS_REJECTING_TEMPERATURE =
+  /^claude-(opus-(4-7|4-8|5)|sonnet-5|fable-5|mythos-5)/;
+
+function anthropicAcceptsTemperature(model: string): boolean {
+  return !ANTHROPIC_MODELS_REJECTING_TEMPERATURE.test(model);
+}
 const DEFAULT_SCHEMA_NAME = 'structured_output';
 const MAX_DETAIL_BODY_CHARS = 500;
 
@@ -548,7 +573,7 @@ function buildAnthropicRequest(ctx: BuildContext): BuiltRequest {
     max_tokens: ANTHROPIC_MAX_TOKENS,
     messages: [{ role: 'user', content }],
   };
-  if (ctx.temperature !== undefined) {
+  if (ctx.temperature !== undefined && anthropicAcceptsTemperature(ctx.model)) {
     body.temperature = ctx.temperature;
   }
   if (ctx.jsonSchema) {
--- a/SparkyFitnessServer/ai/providerDispatch.ts
+++ b/SparkyFitnessServer/ai/providerDispatch.ts
@@ -100,16 +100,9 @@ const OLLAMA_DEFAULT_TIMEOUT_MS = 120_000;
 // response together. At 2048 with a forced tool call, reasoning could consume
 // the budget and truncate the tool response, surfacing as stop_reason
 // 'max_tokens'. 2048 was also tight for a structured nutrition payload even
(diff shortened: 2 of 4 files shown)


CodeWithCJ:
Thanks for putting this together @vaibhav8a! The solution and test coverage look solid, and raising the default headroom to 8192 makes a lot of sense with adaptive thinking.

One request before we merge:

I would like to avoid introducing a new environment variable (`SPARKY_FITNESS_ANTHROPIC_MAX_TOKENS`) for this. In SparkyFitness, I would like env variables are strictly reserved for core system setup (database connections, server ports, core security secrets, etc.) rather than provider-specific integration knobs. This prevents accumulating a sprawl of scattered environment variables that are hard for self-hosters and users to manage or discover. Provider-specific settings should either have sensible built-in defaults or be managed directly via the UI/database.

Since all modern Claude models in our catalog (`claude-sonnet-5`, `claude-opus-5`, `claude-sonnet-4-6`, `claude-haiku-4-5`, etc.) natively support 8192 output tokens, could we simplify this by dropping the env variable logic and using a constant default?

```typescript
const ANTHROPIC_MAX_TOKENS = 8192;
```

If you still prefer a way for customization, it should be via UI. And also, have a look at the coderabbit comment as well.

vaibhav8a:
Thanks — that reasoning makes sense, and it keeps the env surface honest for self-hosters. Dropped the variable and the parsing around it; `max_tokens` is now just the constant:

```ts
const ANTHROPIC_MAX_TOKENS = 8192;
```

The two tests that covered the override and its bad-input fallback are gone with it, since there is no longer anything to override. That also resolves the CodeRabbit note, which was about the parser accepting a value like `8192abc` — the parser no longer exists. The remaining test still asserts the request goes out with 8192 so the headroom can't silently regress.

Server suite passes locally (94 tests in `providerDispatch.test.ts`, 2969 across the server).

CodeWithCJ (approved):

PR #2161, opened 2026-08-18
(fix) Supplements display on Nutrition Summary

### LINKED ISSUE #2145: [Bug]: Supplements not displayed on Nutrition Summary

None

### What happened?

Supplement nutritional payloads do not display on the Diary Nutrition Summary. When logging a supplement with a nutritional payload, the nutrients do not show in the diary.

I checked and it does seem like they are factored into the reports tab.

### Steps to reproduce

1. Create a test Supplement (or a real one)

Image

3. Log the supplement

Image

4. Compare the nutrient values in the summary vs the payload quantity vs the report for that nutrient

Image

Image

### Expected behavior

Nutrient intake should be consistent across diary summary, supplement values, and reporting.

Docker Compose

### SparkyFitness version

1.6.2

### Which area is affected?

_No response_

```yaml

```

```shell

```

_No response_

### Anything else?

_No response_

jsandai:
## Description

A logged supplement dose reaches Reports in full but only partially reached the Diary's Nutrition Summary card. `/api/daily-summary` rolled up five of the seventeen `food_variants` nutrient columns (calories, protein, carbs, fat, dietary fiber) but no custom nutrients, so any other nutrient a supplement carries reads food-only on the Diary while Reports counts the dose. This is a wrong row rather than a missing one: the card already renders every nutrient the user enabled, the number behind it was just short.

Two commits. The first alone leaves the typical supplement understated, because only six of the thirty-two micronutrient catalog entries map to a fixed column (calcium, iron, potassium, sodium, vitamin A, vitamin C); magnesium, D, K, zinc and the B vitamins are all custom nutrients.

1. **Count all fixed supplement nutrients.** Widened the rollup to `FOOD_VARIANT_NUTRIENT_FIELDS`, the list already shared by the report range query and the card's fixed fields, rather than a second hardcoded set that can drift from it. `supplementTotalsSchema`, `EMPTY_SUPPLEMENT_TOTALS`, `getDailySupplementTotals`, the `dailySummaryService` error fallback, web `addSupplementTotals` and mobile `DailyNutritionDetailsScreen` all derive from it.
2. **Count supplement custom nutrients.** `getDailySupplementTotals` now aggregates them by nutrient name (empty map, never null). `supplementCustomRows` is factored out of `supplementCustomUnion` so the supplement-only aggregation and the food union share one SELECT and the status filter and dose clamp cannot drift apart. A shared `addSupplementCustomNutrients` is called by both web and mobile, so one day cannot produce two magnesium figures on adjacent screens.

No migration, no new endpoint, and no display defaults changed: an enabled row simply becomes correct.

Linked Issue: Closes #2145

## How to Test

1. Check out this branch and start the server and frontend as usual.
2. Create a food marked as a supplement carrying a fixed nutrient outside the macros (for example calcium) and a couple of custom nutrients (Magnesium, Vitamin D).
3. Log a dose for today.
4. `GET /api/daily-summary` for that date: `supplementTotals` returns all seventeen fixed keys plus `custom_nutrients` with the dose contribution.
5. On the Diary, enable Calcium in the Summary view, and add Magnesium there as a custom nutrient. Both rows now include the dose and agree with the Reports tab for the same day.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

- [x] **[MANDATORY for Frontend changes] Translations**: no locale files are changed by this PR.

- [x] **[MANDATORY for Backend changes] Code Quality**: typecheck, lint and tests all run clean. No new endpoints; the existing `/api/daily-summary` Zod schema is widened alongside the change.
- [x] **[MANDATORY for Backend changes] Database Security**: no new tables, so `rls_policies.sql` needs no change.

## Screenshots

<details>

2026-08-18 11_33_03-supplements-vitamins - File Explorer

</details>

## Notes for Reviewers

Four behaviour changes worth mentioning:

- **`resolveSupplementTotals` is no longer identity-preserving.** It merges against a full-width zero object, so a client newer than its server still gets seventeen keys. Otherwise `food[key] + supplements[key]` would write `NaN` into the twelve the old server omits.
- **It also always builds a fresh object**, including a fresh `custom_nutrients`. Returning the shared `EMPTY_SUPPLEMENT_TOTALS` would let one caller folding into the result poison the constant process-wide.
- **`hasSupplementNutrition` was silently broken by the widening and is fixed here.** `Object.values(...).some(v => v > 0)` compared the `custom_nutrients` object against `0` and always got false, so a magnesium-only day presented as empty despite a nonzero total. It now checks the two arms separately.
- **Custom values are coerced on both sides**, since `sf_try_numeric` yields NULL for an unparseable value and a key whose every contribution is NULL arrives as JSON null.

Left alone on purpose: the legacy `getDailyNutritionSummary` and its `foodEntryService` fallback, which reports five `total_*` macros plus custom nutrients, has no micronutrient fields to understate, and is not what the Diary calls.

I also considered auto-provisioning supplement nutrients onto the Summary view and decided against it on density grounds: `goal` and `report_*` are scrollable, but the Diary summary is a fixed at-a-glance panel and a multivitamin's many rows would bury the macros. `NutrientDisplaySettings` already offers custom nutrients for the `summary` group, so anyone who wants calcium or D3 there can add it today.

Verification: full gate green on both commits and again after rebasing onto current main (four typechecks, server 3041 pass / 219 skip, frontend 896 pass, mobile 5356 pass, i18n audit clean). Seven mutation controls, each confirmed to land and each killed by a named test, with an unmutated control green.

## Summary by CodeRabbit

* **New Features**
  * Supplement totals now include all supported vitamins, minerals, and custom nutrients—not just core macros.
  * Daily nutrition summaries and details incorporate custom nutrients from supplements.
  * Partial or incomplete supplement data is normalized safely with zero values.

* **Bug Fixes**
  * Fixed supplement-only days being incorrectly shown as empty.
  * Prevented missing nutrient values from producing invalid totals.
  * Preserved existing custom nutrient data without unintended changes.

### DIFF
--- a/SparkyFitnessFrontend/src/tests/utils/addSupplementTotals.test.ts
+++ b/SparkyFitnessFrontend/src/tests/utils/addSupplementTotals.test.ts
@@ -1,4 +1,5 @@
 import { addSupplementTotals } from '@/utils/nutritionCalculations';
+import { EMPTY_SUPPLEMENT_TOTALS } from '@workspace/shared';
 import type { MealTotals } from '@/types/meal';
 
 const foodTotals = {
@@ -11,19 +12,18 @@ const foodTotals = {
   sodium: 2000,
   cholesterol: 100,
   saturated_fat: 10,
+  calcium: 256.3,
   custom_nutrients: {},
 } as unknown as MealTotals;
 
 const supplementTotals = {
+  ...EMPTY_SUPPLEMENT_TOTALS,
   calories: 15,
-  protein: 0,
-  carbs: 0,
   fat: 1.5,
-  dietary_fiber: 0,
 };
 
 describe('addSupplementTotals', () => {
-  it('adds the supplement arm to the five rolled-up fields', () => {
+  it('adds the supplement arm to the macro fields', () => {
     const totals = addSupplementTotals(foodTotals, supplementTotals);
 
     expect(totals.calories).toBe(1547);
@@ -31,14 +31,49 @@ describe('addSupplementTotals', () => {
     expect(totals.protein).toBe(120);
   });
 
-  it('leaves fields the server does not roll up for supplements alone', () => {
-    // Sodium is not summed for supplements by the nutrition summary or the chatbot, so
-    // adding it here would make the Diary the only surface that counted it.
+  // The #2145 case: a supplement's calcium reached Reports, which sums all seventeen
+  // fixed columns, but not the Diary card, which summed five. The reporter saw the
+  // food-only 256.3mg against a dose supplying 10000mg.
+  it('adds micronutrients the supplement carries, not just macros', () => {
+    const totals = addSupplementTotals(foodTotals, {
+      ...EMPTY_SUPPLEMENT_TOTALS,
+      calcium: 10000,
+      sodium: 150,
+      iron: 18,
+    });
+
+    expect(totals.calcium).toBe(10256.3);
+    expect(totals.sodium).toBe(2150);
+    expect(totals.iron).toBe(18);
+  });
+
+  it('leaves a fixed field alone when the supplement contributes nothing to it', () => {
     const totals = addSupplementTotals(foodTotals, supplementTotals);
 
     expect(totals.sodium).toBe(2000);
     expect(totals.sugars).toBe(30);
     expect(totals.cholesterol).toBe(100);
+  });
+
+  // A client newer than its server gets only the five macro keys back. Adding
+  // `food[key] + supplements[key]` across all seventeen would write NaN into the other
+  // twelve, which renders as "NaN mg" rather than failing loudly.
+  it('does not produce NaN against a server that reports only the macro fields', () => {
+    const legacyServerResponse = {
+      calories: 15,
+      protein: 0,
+      carbs: 0,
+      fat: 1.5,
+      dietary_fiber: 0,
+    };
+
+    const totals = addSupplementTotals(foodTotals, legacyServerResponse);
+
+    expect(totals.calories).toBe(1547);
+    expect(totals.calcium).toBe(256.3);
+    expect(totals.sodium).toBe(2000);
+    expect(Number.isNaN(totals.calcium as number)).toBe(false);
+    expect(Number.isNaN(totals.potassium as number)).toBe(false);
   });
 
   it('returns the food totals untouched when there is no supplement arm', () => {
@@ -47,15 +82,10 @@ describe('addSupplementTotals', () => {
   });
 
   it('is a no-op for a day with supplements that carry no nutrition', () => {
-    const totals = addSupplementTotals(foodTotals, {
-      calories: 0,
-      protein: 0,
-      carbs: 0,
-      fat: 0,
-      dietary_fiber: 0,
-    });
+    const totals = addSupplementTotals(foodTotals, EMPTY_SUPPLEMENT_TOTALS);
 
     expect(totals.calories).toBe(1532);
     expect(totals.fat).toBe(50);
+    expect(totals.calcium).toBe(256.3);
   });
 });
--- a/SparkyFitnessFrontend/src/tests/utils/addSupplementTotals.test.ts
+++ b/SparkyFitnessFrontend/src/tests/utils/addSupplementTotals.test.ts
@@ -47,6 +47,59 @@ describe('addSupplementTotals', () => {
     expect(totals.iron).toBe(18);
   });
 
+  // The larger half of #2145. Only six catalog micronutrients have a fixed column, so a
+  // magnesium or vitamin D supplement contributes nothing the fixed loop can reach; those
+  // values live in the snapshot's custom_nutrients and have to be folded separately.
+  it('adds custom nutrients the supplement carries', () => {
+    const totals = addSupplementTotals(
+      {
+        ...foodTotals,
+        custom_nutrients: { Magnesium: 120, Choline: 200 },
+      } as unknown as MealTotals,
+      {
+        ...EMPTY_SUPPLEMENT_TOTALS,
+        custom_nutrients: { Magnesium: 400, 'Vitamin D': 50 },
+      }
+    );
+
+    expect(totals.custom_nutrients).toEqual({
+      Magnesium: 520,
+      Choline: 200,
+      // No food supplied it, which is the ordinary case for a supplement nutrient. Being
+      // dropped for that would leave the row reading zero against a dose that supplied it.
+      'Vitamin D': 50,
+    });
+  });
+
+  it('does not mutate the food custom-nutrient map it was handed', () => {
+    // The caller's own day totals hold this object; folding doses into it in place would
+    // double-count the moment the same day is folded again.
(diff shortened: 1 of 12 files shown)


jsandai:
@coderabbitai

Leaving this one as is.

`/daily-summary` has no 200 response schema for any field today. The route JSDoc documents its 200 as a description string only, and `config/swagger.ts` has no entry for the endpoint at all, so `goals`, `foodEntries`, `exerciseEntries`, `waterIntake` and `calorieBalance` are equally undocumented. `supplementTotals` itself shipped in #2120; this PR widens what it carries rather than introducing it.

So the gap is real but it predates this change and covers the whole response. Documenting one field inside an otherwise undocumented 200 would leave the schema claiming less than the endpoint returns, which is the kind of half-true documentation that's worse than none.

CodeWithCJ:
Claude comments:

1. The fallback hands back the shared mutable constant — [dailySummaryService.ts:305](vscode-webview://0nfha17fj1gjq2spq2tlibua1eomshfoh5cctkn5o2d7msqist1n/SparkyFitnessServer/services/dailySummaryService.ts#L305) returns EMPTY_SUPPLEMENT_TOTALS directly, and that object flows into the response payload. This is the exact aliasing hazard the PR's own notes argue against for resolveSupplementTotals ("would let one caller poison the constant process-wide"). Nothing mutates it today, so it's latent, not a live bug — but { ...EMPTY_SUPPLEMENT_TOTALS, custom_nutrients: {} } costs nothing and closes it.

2. The drift the PR set out to prevent still exists on the Reports side. [reportRepository.ts:630-649](vscode-webview://0nfha17fj1gjq2spq2tlibua1eomshfoh5cctkn5o2d7msqist1n/SparkyFitnessServer/models/reportRepository.ts#L630-L649) has its own hand-written rangeCols list and its own copy of the supplementFixed SQL string. They happen to match FOOD_VARIANT_NUTRIENT_FIELDS today — I checked — but the PR's whole argument is "a second hardcoded set that can drift," and this is that second set. Worth pointing rangeCols at the shared const (it needs an alias map for dietary_fiber → fiber / sugars → sugar) and lifting the duplicated SQL fragment.

3. 17 correlated subqueries where one scan would do. getDailySupplementTotals now emits a separate SELECT SUM(...) FROM medication_entries per nutrient — was 5, now 17, plus the custom aggregation. idx_medication_entries_entry_date (user_id, entry_date) covers it and a day has a handful of rows, so this is not a practical problem. But a single pass (SELECT SUM(sf_try_numeric(...)) AS calories, ... FROM medication_entries WHERE ... once) is simpler and strictly cheaper if you're touching the file anyway.

4. One comment now overstates. The addSupplementTotals docblock keeps "a day with no supplements is byte-for-byte what it was before" — true for the null early return, but when supplementTotals is present-and-empty the result now always carries a custom_nutrients: {} key that food-only totals may not have had. Harmless downstream (dayTotals.custom_nutrients?.[nutrient] ?? 0), just an inaccurate comment.

jsandai:
> Claude comments:
> 
> 1. The fallback hands back the shared mutable constant — dailySummaryService.ts:305 returns EMPTY_SUPPLEMENT_TOTALS directly, and that object flows into the response payload. This is the exact aliasing hazard the PR's own notes argue against for resolveSupplementTotals ("would let one caller poison the constant process-wide"). Nothing mutates it today, so it's latent, not a live bug — but { ...EMPTY_SUPPLEMENT_TOTALS, custom_nutrients: {} } costs nothing and closes it.

**1. Shared constant on the fallback.** Fixed. One note on the fix: `{ ...EMPTY_SUPPLEMENT_TOTALS, custom_nutrients: {} }` works, but a bare spread wouldn't, since it's shallow and `custom_nutrients` would still alias. I've used `resolveSupplementTotals(null)` instead. It already does exactly this, and its docblock is where the argument you're quoting lives, so the code now matches its own comment rather than restating it in a second place.

Added a test that mutates the first degraded response and asserts the second one still comes back zeroed, so it can't quietly regress. Confirmed it fails if I put the constant back.

> 2. The drift the PR set out to prevent still exists on the Reports side. reportRepository.ts:630-649 has its own hand-written rangeCols list and its own copy of the supplementFixed SQL string. They happen to match FOOD_VARIANT_NUTRIENT_FIELDS today — I checked — but the PR's whole argument is "a second hardcoded set that can drift," and this is that second set. Worth pointing rangeCols at the shared const (it needs an alias map for dietary_fiber → fiber / sugars → sugar) and lifting the duplicated SQL fragment.

**2. `rangeCols` in `reportRepository`.** You're right that it's a second hardcoded set. I'd rather not pull it into this PR: `reportRepository.ts` isn't in this diff at all, that duplication shipped with #2120, and the reports path isn't covered by what I've verified end to end here.

Two things that make it a bigger job than it looks:

- The alias map (`dietary_fiber` -> `fiber`, `sugars` -> `sugar`) is itself hand-maintained, so pointing at the shared const shrinks the drift surface without closing it.
- The two `supplementFixed` copies aren't interchangeable. The one in `foodMisc` binds `$1`/`$2`, the one in `reportRepository` correlates to `d.entry_date` in the outer row. Sharing it means parameterising the date expression, which is a real SQL change on the reports query.

Happy to do it, just as its own PR where it can be reviewed as a refactor.

> 3. 17 correlated subqueries where one scan would do. getDailySupplementTotals now emits a separate SELECT SUM(...) FROM medication_entries per nutrient — was 5, now 17, plus the custom aggregation. idx_medication_entries_entry_date (user_id, entry_date) covers it and a day has a handful of rows, so this is not a practical problem. But a single pass (SELECT SUM(sf_try_numeric(...)) AS calories, ... FROM medication_entries WHERE ... once) is simpler and strictly cheaper if you're touching the file anyway.

**3. 17 correlated subqueries.** Also right, and by Claude's own read it isn't a practical problem. Single-pass would work, an aggregate with no `GROUP BY` returns a row even with no matches and the `Number(row[field]) || 0` coercion already handles the NULLs.

> 4. One comment now overstates. The addSupplementTotals docblock keeps "a day with no supplements is byte-for-byte what it was before" — true for the null early return, but when supplementTotals is present-and-empty the result now always carries a custom_nutrients: {} key that food-only totals may not have had. Harmless downstream (dayTotals.custom_nutrients?.[nutrient] ?? 0), just an inaccurate comment.

**4. Comment overstates.** Agreed, reworded. It's actually a bit worse than "present-and-empty": `supplementTotals` is non-nullable on the response, so against this server the early return only fires for an older server or a failed fetch, never for a supplement-free day. The old sentence described a path that basically doesn't happen anymore. And you're right that it's harmless downstream, `custom_nutrients?.[nutrient] ?? 0` reads an empty map and an absent one the same way.

Worth stating that 2 and 3 pull opposite ways though: 2 says share the `supplementFixed` fragment more widely, 3 says stop using it in the one place that does. A single-pass version needs a differently shaped helper, so I'd like to do them together rather than have one undo the other.

I'll open the follow-up once this lands.

CodeWithCJ (approved):

Issue #2229, opened 2026-08-24
[Feature]: Expand MCP / chatbot tool registry so AI agents can manage SparkyFitness autonomously


Rahulsharma0810:
**Are you willing to submit a PR?** Yes, I'd like to submit a PR!

## What problem does this solve?

The MCP server (and the in-process chatbot) exposes only a subset of SparkyFitness's capabilities. The MCP surface is generated from the tool registry in `SparkyFitnessServer/ai/tools/index.ts` (`buildChatbotTools(..., providerTuning=false)`, consumed by `ai/mcp/mcpAdapter.ts`), which only composes **9 categories**: `exercise, food, checkin, goals, coaching, vision, profile, reports, medications`.

Meanwhile the REST API (`/api/api-docs/json`) exposes **421 operations across 21 tag groups**. Many everyday features therefore can't be driven by an AI agent over MCP, so an agent can't manage an account end-to-end.

## Describe the feature you'd like

Add tool builders for the **non-security-critical** domains below so they're reachable via MCP and chat. Each would follow the existing contract documented in `SparkyFitnessServer/AGENTS.md`: flat Zod schema → strict union `safeParse` inside `execute` → orchestrate existing services/repositories → return `ERRORS.*` strings → golden tests in `tests/chatbotTools*.test.ts`.

**Explicitly out of scope (security-sensitive):** admin user/role/status management, backups, OIDC settings, MFA, API-key generation, email/password changes, family-access *mutation*, switch-context.

Proposed new/extended tools (missing API surface → proposed action):

| Domain | Missing capability | Proposed tool / action |
|---|---|---|
| Nutrition | barcode lookup, favorites, meal-plan templates, custom nutrients, meal-types CRUD, CSV export | extend `food`; new `sparky_manage_meal_plans`, `sparky_manage_favorites`, `sparky_manage_custom_nutrients` |
| Fitness | workout plan templates, log-from-plan, grouped sessions, PR matrix, exercise stats/summary | extend `exercise`; new `sparky_manage_workout_plans`, `sparky_get_exercise_stats` |
| Wellness | water containers, progress/check-in photos (read), fasting stats/history-range, sleep analytics | extend `checkin` |
| SleepScience | chronotype, sleep debt, MCTQ baseline, energy curve, daily need | new `sparky_sleep_science` |
| Goals | adaptive TDEE, goal presets, weekly goal plans | extend `goals` |
[...]

PR #2230, opened 2026-08-24
feat(ai): addition of chatbot/MCP tools and extend medication management

Rahulsharma0810:
## Summary

Closes part of #2229. Expands the SparkyFitness chatbot / MCP tool registry so an AI agent can manage SparkyFitness autonomously, exposing previously web-only domains as first-class tools. Every domain follows the existing tool contract (flat Zod schema, strict discriminated-union validation inside `execute`, never throws — returns `ERRORS.*` strings, golden-tested exact output).

Security-critical surfaces are deliberately **excluded**: admin/user management, MFA, API-key management, OIDC/SSO config, backups, family-access mutation, and integration credential/token management.

## Tools added / extended (14 domains, 15 commits)

| Tool | Category | Actions |
| --- | --- | --- |
| `sparky_manage_allergens` | allergens | list / add / remove |
| `sparky_manage_medications` *(extended)* | medications | + create / update / delete medication, list / add / delete schedule (14 actions total) |
| `sparky_manage_favorites` | favorites | list / add / remove (food & meal) |
| `sparky_manage_meal_plans` | meal_plans | list / get / duplicate / delete |
| `sparky_manage_custom_nutrients` | custom_nutrients | list / get / create / update / delete |
| `sparky_manage_water_containers` | water_containers | list / get / create / update / delete / set primary |
| `sparky_manage_workout_plans` | workout_plans | list / get / delete |
| `sparky_get_exercise_stats` | exercise_stats | stats summary / query activities / personal records / matched courses |
| `sparky_get_sleep_science` | sleep_science | sleep debt / MCTQ stats / daily need / energy curve / chronotype / data sufficiency / recalculate baseline |
| `sparky_get_integrations` | integrations | list providers / list provider types (credential-safe, read-only) |
| `sparky_manage_synced_data` | synced_data | list synced sources / delete synced source |
| `sparky_manage_progress_photos` | progress_photos | list photo dates / list photos / delete photo |
| `sparky_get_dashboard` | dashboard | daily calorie-balance summary |
| `sparky_get_barcode` | barcode | lookup food by barcode across saved foods + external providers |

## Implementation notes

- Each new category is wired through the full registry: shared `CHAT_TOOL_CATEGORY_SLUGS`, per-domain schema + builder, `ai/tools/index.ts` (`CATEGORY_BUILDERS` + `CATEGORY_ORDER`), `metaTools` category summaries, `chatService` keyword routing, plus the frontend category selector + `en` translations.
- Read-only analytics tools (`sparky_get_*`) never mutate; mutation tools reuse existing services/repositories under the caller's RLS context.
- Credential/token fields are stripped from integration output; only non-secret status is surfaced.

## Testing

- Per-domain golden tests assert exact tool output (parity contract with the MCP surface).
- `pnpm run validate` (typecheck + eslint + prettier) clean on server and frontend.
- Full server suite: **3327 passed | 221 skipped**.
- Frontend `ChatToolCategories` tests pass.

---
## ✅ Required Checklist (restored automatically)

> has been added back below. **Please tick each box and do not delete this section** —
> no new commit is needed.

- [ ] **[MANDATORY for UI changes] Screenshots**: I have attached Before/After screenshots below. _(N/A — no UI changes; only backend tools + two i18n label additions to the English translation file and the chat category selector list.)_

## Summary by CodeRabbit

* **New Features**
  * Expanded chatbot capabilities for allergens, favorite foods, meal plans, custom nutrients, water containers, workout plans, exercise analytics, sleep insights, integrations, progress photos, dashboards, and barcode lookups.
  * Added support for medication and schedule management through the chatbot.
  * Added pagination for allergen lists and richer formatted summaries.

* **Bug Fixes**
  * Improved not-found handling, validation, error messaging, unit conversions, date defaults, and case-insensitive synced-data matching.
  * Synced-data chatbot access is now read-only; deletion remains available in the web interface.


CodeWithCJ:
Thanks for your PR. please take a look at the bot feedbacks and also complete mandatory questionnaire in the PR desc.

CodeWithCJ:
Could you look into the new bot comments one more time

CodeWithCJ:
A few things to fix before this can go in.

### 1. The allergen keyword rule pulls food logging away from the food tools

The `allergens` rule in `KEYWORD_RULES` matches bare food nouns — `peanut\w*`, `dairy`, `gluten`, `lactose`, `shellfish`, `soy`, `sesame`, `tree\s*nut\w*`. `classifyUserIntent` returns as soon as the keyword classifier matches anything, so these never reach the LLM fallback:

| message | `main` | this branch |
| --- | --- | --- |
| `log 2 tbsp of peanut butter` | `[]` → LLM picks food | `[allergens]` |
| `add a glass of dairy milk` | `[]` → LLM picks food | `[allergens]` |
| `log soy sauce with my rice` | `[]` → LLM picks food | `[allergens]` |
| `log 30g of sesame seeds` | `[]` → LLM picks food | `[allergens]` |
| `had a lactose free yogurt` | `[]` → LLM picks food | `[allergens]` |

No food tool gets loaded, so the chatbot can't log the meal. Match intent words instead:

```
/\b(allerg\w*|intoleran\w*|anaphyla\w*)\b/i
```

If the food nouns should still count, have them add `food` alongside `allergens` rather than replacing it.

### 2. `sparky_manage_medications` action inference

`hasMedicationUpdateField` runs before the dose-logging checks and includes `notes`, which `log`, `log_injection` and `update_entry` all already use:

| call | `main` | this branch |
| --- | --- | --- |
| `{medication_id, notes}` | `get_medication` | `update_medication` — silently overwrites the medication's notes |
| `{medication_id, dose_mg, site, notes}` | injection logged | `VALIDATION: Unrecognized keys: "dose_mg", "site"` |
| `{medication_id, status:'taken', notes}` | dose logged | `VALIDATION: Unrecognized key: "status"` |

"Log my dose, felt nauseous" rewrites the medication record instead of logging anything.

Move the block below the `entry_id` / `site` / `dose_mg` / `status` checks, or drop `notes` from the field set and require an explicit action for medication mutations — `favoritesTools.ts` already does the latter for this reason.

### 3. Water container volumes print in ml against the user's unit

`volume` is persisted in ml while `unit` keeps what the user picked, so `formatContainer` mislabels anything not in ml. A 24 oz bottle renders as:

```
**Hydro Flask** 709.764oz — primary
```

Use the conversion `WaterContainerManager.tsx` already does (`convertMlToSelectedUnit`). Add an `oz` case to `chatbotToolsWaterContainers.test.ts` — every case there is `unit: 'ml'`, which is why this passes.

### 4. `update_water_container` accepts `volume` without `unit`

`waterContainerService.updateWaterContainer` logs a warning and stores the raw number as ml, so "make my bottle 24 oz" can become 24 ml. Add a `.refine()` requiring `unit` whenever `volume` is present.

### 5. Deletes report success unconditionally

`checkInPhotoService.deletePhoto` returns early when nothing matches, but `delete_photo` still returns "Progress photo deleted." `delete_meal_plan` ignores its result the same way. Both should return `NOT_FOUND`, like the allergen tool does.

### 6. `delete_meal_plan` description is incomplete

`deleteMealPlanTemplate` also deletes food entries created from that template from today forward. That needs to be in the description — the model reads it to decide whether to call.

### 7. Dead entry in the classifier

`'allergens'` was added to `validCategories` in `chatService.ts`, but `classificationPrompt` doesn't list it as an available domain, so the model can never emit it. Add it to the prompt or drop it.

### 8. Please trim the category list back

The 13 new categories are all single-tool, while the existing 9 bundle whole domains (food 8 tools, exercise 9, coaching 9). Several of the new ones are subsets of categories that already exist:

- `food`'s summary is "log meals/water" — Water Containers is now a sibling
- `progress_photos` reads the `check_in_photos` table but sits outside Body & Check-ins
- `dashboard` is a daily summary; `reports` is already "daily/weekly summaries"
- `barcode`, `favorites`, `meal_plans`, `custom_nutrients`, `allergens` are all food
- `exercise_stats`, `workout_plans` are exercise

Could you fold these into the existing domains so the selector comes back to roughly its previous size? It went from 9 chips to 22, and the two levels of granularity are hard to tell apart in the UI.

Folding them in also removes the root cause of item 1 above, since food keywords would no longer pull the chat away from the food tools.

If any genuinely need to stay top-level, the selector needs grouped sections to stay readable at that count — it currently renders one flat `flex flex-wrap` over the raw slug array.

CodeWithCJ:
Follow-up after exercising these tools against a live instance. Five more items, continuing the numbering from my previous comment.

### 9. `sparky_manage_workout_plans` — `get_workout_plan` and `delete_workout_plan` cannot be called at all

`workout_plan_templates.id` is `SERIAL PRIMARY KEY` (integer, from `20251008224256_add_workout_features.sql`), but `schemas/workoutPlans.ts` types `plan_id` as `uuidSchema` for both actions.

`list_workout_plans` renders an integer ID, and feeding that same ID back gives:

```
Error [VALIDATION]: plan_id: Must be a valid UUID
```

There is no input that satisfies it, so 2 of the tool's 3 actions are dead. The existing REST route never required a UUID — it passes `req.params.id` straight through.

Use the same numeric ID schema you already used for water containers:

```ts
const planIdSchema = z.coerce
  .number()
  .int()
  .positive('Workout plan ID must be a positive integer');
```

`chatbotToolsWorkoutPlans.test.ts` uses `PLAN_ID = '123e4567-e89b-12d3-a456-426614174000'`, a UUID that can never exist in an integer column, plus a case asserting non-UUIDs are rejected — which locks the bug in. Those fixtures need to become integers.

I checked the ID type of every table the new tools touch; this is the only mismatch.

### 10. `sparky_get_sleep_science` renders a raw null

`mctq_stats` prints the social jetlag field with no null guard, so when it isn't set the output is:

```
- Social jetlag: null h
```

Guard it the way `exerciseStatsTools.ts` already guards `avgHeartRate` (`?? 'n/a'`).

### 11. `sparky_get_exercise_stats` prints an unrounded float

`stats_summary` emits the raw duration, e.g.:

```
- Duration: 1429.5634166666666 min
```

Every other figure in that block is clean. Round it.

### 12. `sparky_get_exercise_stats` header is misleading

The heading renders as `Exercise Stats (month: <start> → <end>)`, but `interval: 'month'` means "bucket by month" and the service defaults to a **six month** window. Reading that header, the model will summarise six months of training as "this month", and the period-over-period percentages get attributed to the wrong window too. The service behaviour is fine — the wording isn't. Something like `bucketed by month` alongside the explicit range would fix it.

### 13. `delete_synced_source` is case-sensitive, and real source values differ only by case

Source strings that differ only in capitalisation are stored as separate values, so a single `delete_synced_source` call clears one and silently leaves the other. Combined with item 5 (the tool reports success unconditionally), the model will tell the user everything was deleted while a substantial number of rows remain.

Since this is a destructive, LLM-invoked bulk delete, it should either match case-insensitively, or `list_synced_sources` should merge case variants so the model can't pick just one. Worth reporting the actual deleted count back in the confirmation too.

Rahulsharma0810:
@CodeWithCJ Thanks for the review! Both items are now addressed:

**Mandatory questionnaire** — completed in the PR description (Integrity & License, Frontend Quality, Frontend Translations, Backend Code Quality, and Alignment all checked; UI Screenshots N/A as this is a backend/tooling-only change).

**Bot feedback** — all CodeRabbit findings resolved across two passes:
- First pass (`3f4beff7`): fixed inferAction edge cases (favorites, medications), added `.strict()` to discriminated-union schemas, required ≥1 field in update schemas, conditional requirements in medication schedules, test mock `.js` specifiers, frozen clocks in date-default tests, and NOT_FOUND propagation for missing workout plans.
- Second pass (`1f6f7c5d`): empty-page pagination footer, fallback classification now covers every registered tool category, and `update_medication` passes `null` through so fields can be cleared.

All review conversations are resolved and every required check is green (Validate & Label, Server Tests, Frontend Tests, CodeRabbit, GitGuardian). Ready for your review whenever you have a moment. 🙏

CodeWithCJ:
@Rahulsharma0810 The CodeRabbit passes and the questionnaire look done, but the two review comments I left above are separate from the bot findings and are still open — items 1 through 13.

Checked against 1f6f7c5. The `validCategories` change touches item 7 without closing it: the accept-list now takes all 22 categories, but `classificationPrompt` still lists 7 domains, so the model is never told the other 15 exist. Same mismatch, just wider.

Still reproducing on this head:

```
"log 2 tbsp of peanut butter"          => [allergens]           item 1, no food tool loaded
{medication_id, notes}                 => "Medication updated"  item 2, silent overwrite
{medication_id, site, dose_mg, notes}  => Unrecognized keys: "site", "dose_mg"
{medication_id, status, notes}         => Unrecognized key: "status"
```

Unchanged in code: items 3, 4, 5, 6, 8, 9, 10, 11, 12, 13.

The suite being green doesn't clear these — it passes because the fixtures encode the bugs. `chatbotToolsWorkoutPlans.test.ts` asserts a UUID `plan_id` against an integer column, and every water container fixture is `unit: 'ml'`. Both need updating alongside the fixes.

Main blockers are 1, 2, 3 and 9.

Rahulsharma0810:
Thanks for the detailed review, @CodeWithCJ. I've pushed two commits that address all 13 items — `3ebbd00b` (correctness) and `0c8445fb` (the category-fold refactor). Fixtures were updated alongside each fix so the green suite reflects the corrected behaviour rather than the old bugs.

**Blockers**
1. **Allergen keyword rule stealing food logging** — narrowed the allergens intent regex to `/\b(allerg\w*|intoleran\w*|anaphyla\w*)\b/i` so bare food nouns (peanut, dairy, soy, sesame…) no longer classify as allergens. The category fold (item 8) also removes the root cause: allergen keywords now live inside the `food` domain.
2. **`sparky_manage_medications` action inference** — dropped `notes` from the update-field set and moved the update/create detection *below* the `entry_id` / injection / dose-logging / status checks, so `{medication_id, notes}`, `{…, dose_mg, site}`, and `{…, status}` route to the right actions instead of silently updating.
3. **Water container volumes printed as ml** — added `convertMlToUnit` + `formatVolume` mirroring `WaterContainerManager.tsx`, so a 24 oz bottle now renders `24 oz`. Added an `oz` fixture to the golden test (previously every case was `ml`).
9. **Workout plan IDs typed as UUID** — `plan_id` for `get_workout_plan`/`delete_workout_plan` is now `z.coerce.number().int().positive()` to match the SERIAL column; fixtures switched from UUIDs to integers.

**Other items**
4. `update_water_container` now requires `unit` whenever `volume` is present (schema `.refine`).
5. `deletePhoto` returns a boolean and both `delete_photo` and `delete_meal_plan` return `NOT_FOUND` when nothing was removed.
6. `delete_meal_plan` description now states it also deletes food entries created from the plan from today forward.
7. The classifier's Available-domains prompt now lists all nine categories, matching the accept-list.
8. Folded the 13 single-tool categories back into the nine original domains (favorites/meal_plans/custom_nutrients/water_containers/allergens/barcode → food; exercise_stats/workout_plans → exercise; progress_photos/sleep_science → checkin; dashboard → reports; integrations/synced_data → profile). The selector returns to a compact set; the full tool set is unchanged.
10. `mctq_stats` social jetlag now prints `n/a` instead of `null`.
11. Exercise stats duration is rounded.
12. Exercise stats header now reads `bucketed by <interval>` to avoid implying "this month".
13. `delete_synced_source` matches case-insensitively and `list_synced_sources` merges case variants; the confirmation already reports the actual deleted count.

`pnpm run validate` + `pnpm test` are green on both packages (server 3334 passed, frontend incl. ChatToolCategories). Happy to adjust anything further.

CodeWithCJ:
@Rahulsharma0810 All 13 items from my earlier comments check out — verified each one behaviourally, not just by reading the diff. The water container volumes, the workout plan integer IDs, the medication routing and the keyword folding all do the right thing now. Thanks for turning that around.

I've pushed one commit to the branch myself (`f24a7a9`) rather than send it back to you for another round, since it's a scope decision rather than a defect and I didn't want to hold the PR on it.

**What changed:** `delete_synced_source` is removed from the AI tool surface, and `sparky_manage_synced_data` is renamed `sparky_get_synced_data` to match the read-only naming convention.

**Why:** the delete takes only `source` — no date bound, no limit, no dry run. So a request like "remove last week's duplicate garmin steps" has no correct tool call; the model's only option is to delete everything that provider ever wrote, across all ten synced tables, with no undo. And because the tool defines an `execute` function, the SDK runs it the moment the model emits the call — there's no approval step between the decision and the DELETE. The "confirm the source with the user first" line in the description is advisory, and nothing checks that the model honoured it.

`sparky_ask_user` doesn't cover this either: `includeAskTool` defaults to false and is never set on the MCP path.

**What did not change:** `syncedDataService.deleteSyncedSource`, `syncedDataRepository.deleteSyncedDataBySource` and `DELETE /synced-data/:source` are all untouched, along with their tests. The web UI keeps full delete capability with its confirmation dialog and visible counts. This removes the LLM's unsupervised access to the operation, not the operation.

The case-insensitive matching and the `byTable` accumulation fix you added are both still in place and still covered — those were good catches on the listing side.

Two tests were added asserting the delete is rejected and that a stray `source` field can't trigger one, so it can't quietly come back. Server suite is green at 3551 passing, typecheck and lint clean.

Happy to revert this if you feel strongly — but if you want the delete available to the AI, I'd want it to require an explicit confirmation argument plus date bounds first, so a single inferred call can't do it and partial requests have a real answer.

CodeWithCJ:
@coderabbitai resume

CodeWithCJ (approved):

PR #2272, opened 2026-08-27
feat(food-search): allow "All Providers" as the default food provider

### LINKED ISSUE #2263: [Feature]: Allow "All Providers" to be selected as the default food search provider

None

In SparkyFitness 1.6.3, the All Providers option works well when searching for foods across multiple configured sources, but I have to manually select it each time the Add Food dialog opens. For example, if I search for cottage cheese, the default provider may only return Tandoor results even though USDA, FatSecret, or OpenFoodFacts may have better matches. This adds an extra step to every food search.

It would be helpful if All Providers could be saved as the default food search mode, similar to how individual providers can be selected as the default. This is related to #1696, but this request is specifically about allowing the existing All Providers aggregate search mode to be selected and persisted as the default rather than choosing one active provider. 

For example, I have USDA, FatSecret, OpenFoodFacts, and Tandoor connected, and I would prefer each new food search to query all configured providers by default.

Nutrition tracking

Nice to have

The current workaround is to manually select All Providers each time I open the Add Food dialog. This works, but adds an unnecessary extra step to every food search.

### Additional context

All Providers itself is already working well and returning results across my configured sources. This request is only to allow that existing search mode to persist as the default.

Image
Image

jsandai:
## Description

The aggregated "All Providers" food search can only be picked per session, so every new search screen resets to a single provider. This lets it be saved as the default.

New `food_search_all_providers_default` boolean on `user_preferences`, offered in the default-food-provider picker on web and mobile. It's a separate column rather than a value in `default_food_data_provider_id` because that column is a `uuid` and can't hold the `__all__` sentinel. Keeping them apart also means turning the aggregated default off restores the provider the user had picked.

Both clients ignore the preference below two active providers, matching the search dropdown, and fall through to the stored single provider without clearing it. Default is `FALSE`, so out-of-box behaviour is unchanged.

Linked Issue: Closes #2263

## How to Test

1. Check out this branch, run the migration, and start the stack.
2. Configure at least two active food providers.
3. Web: set the default food data provider to "All Providers", open food search, confirm it starts aggregated.
4. Mobile: Food Settings, Default Food Provider, "All Providers"; open search, type 3+ characters, confirm the online header names the aggregated provider.
5. Deactivate the provider you did not store as the default. The aggregated option stops being offered and both pickers show the remaining provider.
6. Now deactivate the stored provider instead. Both pickers fall back to their placeholder rather than showing a stale provider.
7. Re-activate, confirm the aggregated default returns without re-selecting.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

2026-08-26 15_41_02-RustDesk

2026-08-26 15_41_37-Network - File Explorer

Screenshot_20260826_171328_SparkyFitness

Screenshot_20260826_171349_SparkyFitness

</details>

## Notes for Reviewers

One behaviour change beyond the feature itself, flagged because it touches the existing picker: if your stored default provider is no longer active, the settings picker now shows the first active provider instead of a blank. That matches what search already does via `resolveFoodProviderId`, so the two no longer disagree about which provider is in play. A default that was never set still shows the placeholder.

The upsert's `ON CONFLICT` branch reads the positional parameter directly instead of `EXCLUDED`. The `VALUES` clause defaults the column to `false`, so `EXCLUDED` is never NULL and any unrelated upsert that omits the field would otherwise reset a stored `true`. Same shape as `time_format` in that statement.

This PR also unifies the mobile wording on "provider", which web already used. Mobile said "source" in the same places, so `foodSearch.menu.allSources`, `foodSearch.sections.bySource` and `foodSearch.accessibility.sourceChange` are renamed. Only the English catalog is edited, so those three keys show as missing on `pl` and fall back to English until a translator picks them up.

The Database Security box is ticked on the basis that this adds a column to `user_preferences`, which already has row-level policies, and introduces no new table. Say the word if you want anything added to `rls_policies.sql` regardless.


CodeWithCJ:
I was going to tag you to that item and you already implemented it 😄  Thanks!!!

jsandai:
@coderabbitai

Two fixed, two I'd push back on.

**Settings picker fallback: fixed.** Search already falls back to the first active provider via resolveFoodProviderId, so settings showing a blank trigger was telling the user something different from what the app would actually do. Both clients now fall back to the first active provider when the stored one is gone, and stay empty when nothing was ever chosen, so the placeholder still means "nothing picked".

**any in the picker mock: fixed.** Swapped for a small typed shape. Worth noting renderScreen a few lines down still takes any, but that's existing code I didn't touch, so I've left it alone.

**NOT NULL migration lock: leaving as is.** Fair point in general, but user_preferences is one row per user, so even a big self-hosted instance is a few thousand rows and the scan is milliseconds. A CHECK NOT VALID plus VALIDATE dance would add moving parts for no real gain here. Happy to switch if you'd rather every migration follow that pattern regardless of table size.

**any in preferenceRepository: agree, but not here.** Both methods have taken userId: any and preferenceData: any since long before this PR, and I've only added one field. Typing them properly means an interface covering all 47 preferences plus the presence flags, which is a real improvement but a much bigger diff than this feature warrants and would bury the actual change.

Full suite green on the update.

SparkyFitnessServer/db/migrations/20260821170000_add_food_search_all_providers_default.sql
CodeWithCJ, line 1: Could you merge both the SQL into a single file

CodeWithCJ (commented):
Frontend Tests is red on `shared/src/utils/foodEntryCopyFingerprint.ts:103`,
   which is a main issue, already fixed there. Pull main in and re-run: `tsc -b`
   bailed at the shared project, so the frontend typecheck, lint and vitest
   suite never actually ran against your changes.

Non-blocking:

- `SparkyFitnessFrontend/src/utils/settings.ts:283` — showing the first active
  provider when the stored default is inactive is a behaviour change to the
  existing settings picker, not part of this feature. It's fine and it does
  match search, but call it out in the description.
- `SparkyFitnessFrontend/src/pages/Settings/ExternalProviderSettings.tsx:180`
  drops the `defaultValue` that the sibling call in FoodSearch.tsx:1281 passes.
- Keep the source→provider rename in this PR; splitting it isn't worth a second
  round trip.

jsandai:
Rebased onto main and pushed as `98349c3a`.

You were right that nothing was actually running. My branch was 65 commits behind, so tsc -b bailed at the shared project and the frontend checks never got as far as my code. That file didn't even exist at my base. With main pulled in the frontend typecheck, lint and vitest all run and pass, and the server and mobile suites picked up about 285 new tests from the commits I'd been missing, all green.

Migrations merged into the single file, with NOT NULL inline on the ADD COLUMN. My reason for splitting them was that the runner keys on filename, so an amended file never re-runs on a database that already recorded it. That only matters for my own dev box though, since this branch hasn't merged anywhere, so one file is the right call for everyone else.

Added the missing defaultValue so it matches the FoodSearch call.

Called out the settings picker behaviour change in the description.

Keeping the rename here as you suggested.

CodeWithCJ (approved):

PR #2284, opened 2026-08-28
fix(mobile): keep chat scroll position while reading history

### LINKED ISSUE #2276: [Bug]: Android App Sparky AI

None

### What happened?

On android, within the Sparky Fitness app, when trying to scroll back through Sparky AI chat history: 

It is almost impossible to scroll back through Sparky AI chat.  It seems like the app is trying to auto scroll to the end, or most recent, chat inquiry.

Short screen recording attached for clarification.

### Steps to reproduce

Sparky Fitness Android app. Tap on "Ask Sparky". Try to scroll back through previous Sparky AI inquiries and responses. See bug (auto scrolls to most recent response regardless of user input to scroll back further).  Issue persists after a clean install of the app.

### Expected behavior

Expected to be able to scroll back through conversations and stop at any given response to read it without the app auto returning to most recent response.

Docker Compose

### SparkyFitness version

1.6.3(1)

### Which area is affected?

Mobile app (Android)

```yaml

```

```shell

```

### Anything else?

If logs are needed, please advise on easiest export method.  Either docker logs or android logs (logcat perhaps?) can be sent.

thaildhe172591:
> [!TIP]
> For **Frontend** changes, please run `pnpm run validate` to check for any errors.
> PRs that include tests and clear screenshots are highly preferred!
> Note: AI-generated descriptions must be manually edited for conciseness. Do not paste raw AI summaries.

## Description

Scrolling back through Sparky chat history on Android is close to impossible: the list pulls itself back to the newest message. Reported in #2276.

`ChatThread` wired `scrollToBottom` straight to the message list's `onContentSizeChange` and `onLayout`, so any re-measure scrolled to the end regardless of where the user was. The bubbles have variable height and no `getItemLayout`, so FlatList re-measures rows as they come into view, and every row that settles while you scroll back fires the handler and pulls the list down again. `onLayout` does the same on a keyboard show/hide.

The list now tracks whether the user is at the bottom (`onScroll`, 40px threshold) and only auto-scrolls when they are, so mount-time seeding and stream-following are unchanged. Starting a run re-pins to the bottom, which covers send and retry alike: the reply lands at the end and that is what the user just asked to see.

Linked Issue: Closes #2276

## How to Test

1. Check out this branch, run `pnpm install`, then `pnpm run android` in `SparkyFitnessMobile/`.
2. Open Ask Sparky on a thread with enough history to scroll (roughly 20 exchanges).
3. Scroll back toward the older messages and stop. The list stays where you left it.
4. Scroll back to the bottom and send a message: the thread still follows the streaming reply.
5. Scroll up again, then send: the list returns to the bottom so the new reply is visible.

Automated: `pnpm exec jest --watchman=false --runInBand __tests__/screens/ChatScreen.test.tsx`

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

Not attached. See the note below on what I could and could not verify.

</details>

## Notes for Reviewers

Two checkboxes are deliberately left unticked rather than ticked on trust, so please read this before running CI.

I did not get the change onto a running screen. The debug APK builds and installs on an Android 36 emulator here, but the emulator runs on a software GPU and I could not drive it as far as the chat screen, so I have no before/after recording and cannot claim device verification. A scroll bug also needs a recording rather than a still image to show anything useful. If you would rather have that evidence before reviewing, say so and I will get a physical device on it.

What is verified, all on this machine: `pnpm run typecheck` and `pnpm run lint` are clean, and the `ChatScreen` suite is 16/16. The two new pinning tests fail against the current `main` behaviour and pass with this change, so the regression they describe is real and covered. The full mobile suite is 5986 passing; the 18 failures are the same ones present on an untouched checkout of `main` (`workoutSession`, `medicationFormat`, `i18nAudit`, `widgetResourceContract`, `healthDataDisplay`) and are unrelated to this change.

Two smaller decisions worth flagging. The threshold is 40px rather than exact equality because `contentOffset` lands fractionally short of `contentSize - layoutMeasurement` on Android and momentum scrolling stops a few pixels early. Re-pinning is driven by the thread's running state instead of the `composer.send` event because assistant-ui marks that event deprecated in favour of state observation, and the running state covers retry as well as send.

## Summary by CodeRabbit

- **New Features**
  - Improved chat auto-scrolling to keep the latest messages visible while streaming.
  - Preserves the user’s scroll position when they browse older messages.
  - Automatically resumes following new messages when the user returns to the bottom.
  - New messages and retries re-pin the chat to the latest content.

- **Tests**
  - Added coverage for pinned scrolling, manual scrolling, content resizing, and new message runs.

### DIFF
--- a/SparkyFitnessMobile/__tests__/screens/ChatScreen.test.tsx
+++ b/SparkyFitnessMobile/__tests__/screens/ChatScreen.test.tsx
@@ -367,6 +367,124 @@ describe('ChatScreen thread', () => {
   });
 });
 
+describe('ChatScreen auto-scroll pinning', () => {
+  const scrollEvent = (offsetY: number, contentHeight: number, viewportHeight = 600) => ({
+    nativeEvent: {
+      contentOffset: { x: 0, y: offsetY },
+      contentSize: { width: 390, height: contentHeight },
+      layoutMeasurement: { width: 390, height: viewportHeight },
+    },
+  });
+
+  let scrollToEnd: jest.Mock;
+  let animationFrames: FrameRequestCallback[];
+  let requestAnimationFrameSpy: jest.SpyInstance;
+  let cancelAnimationFrameSpy: jest.SpyInstance;
+
+  const flushAnimationFrames = async () => {
+    await act(async () => {
+      while (animationFrames.length > 0) {
+        const callbacks = animationFrames.splice(0);
+        callbacks.forEach((callback) => callback(0));
+      }
+    });
+  };
+
+  beforeEach(() => {
+    scrollToEnd = jest.fn();
+    animationFrames = [];
+    (global as any).__mockChatIsEmpty = false;
+    (global as any).__mockMessagesScrollToEnd = scrollToEnd;
+    requestAnimationFrameSpy = jest
+      .spyOn(global, 'requestAnimationFrame')
+      .mockImplementation((callback) => {
+        animationFrames.push(callback);
+        return animationFrames.length;
+      });
+    cancelAnimationFrameSpy = jest
+      .spyOn(global, 'cancelAnimationFrame')
+      .mockImplementation(() => undefined);
+  });
+
+  afterEach(() => {
+    requestAnimationFrameSpy.mockRestore();
+    cancelAnimationFrameSpy.mockRestore();
+  });
+
+  const tree = (queryClient: QueryClient) => (
+    <QueryClientProvider client={queryClient}>
+      <SafeAreaProvider initialMetrics={initialMetrics}>
+        <ChatScreen navigation={navigation} route={route} />
+      </SafeAreaProvider>
+    </QueryClientProvider>
+  );
+
+  async function renderThread() {
+    const queryClient = new QueryClient();
+    const screen = render(tree(queryClient));
+    const messages = await screen.findByTestId('thread-messages');
+    await flushAnimationFrames();
+    scrollToEnd.mockClear();
+    return { ...screen, messages, queryClient };
+  }
+
+  it('follows content growth while the user is at the bottom (streaming)', async () => {
+    const { messages } = await renderThread();
+
+    fireEvent(messages, 'scroll', scrollEvent(1400, 2000));
+    fireEvent(messages, 'contentSizeChange', 390, 2100);
+    await flushAnimationFrames();
+
+    expect(scrollToEnd).toHaveBeenCalledWith({ animated: false });
+  });
+
+  it('does not snap back to the bottom when content resizes while the user is scrolled up (#2276)', async () => {
+    const { messages } = await renderThread();
+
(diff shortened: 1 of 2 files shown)


CodeWithCJ:
@jsandai  Could you test this on a real device if feasible .

jsandai:
> @jsandai Could you test this on a real device if feasible .

Before testing the fix I attempted to replicate the issue on my device (s26) running 1.6.4 and was unable to do so.

Scrolling back through ai chat history, I can freely navigate and stop anywhere in the transcript - I don't experience any drift or snap to bottom behaviour. Scrolling up and sending a message pulls the stream into focus, that is - once a message is sent it becomes the focus at the bottom of the transcript. This is arguably expected behaviour given a user sending a message would probably expect to see that message followed by its response rather than being held wherever they'd scrolled to within the transcript. The only time the bottom message is forced into view from elsewhere within the transcript is when opening or closing the keyboard but again, this is somewhat defensible, at least in the 'show keyboard' scenario given the user is about to type and send? This PR removes that behaviour - after the fix, opening the keyboard while scrolled up leaves you where you are. I'd say this is a UX call and not necessarily a bug fix.

Happy to build the branch and test it anyway if that's still useful but the fact I'm unable to reproduce the probem means I'll be testing the keyboard case only.

CodeWithCJ:
Could you create a APK in your repo to see if that we can send to the user who raised the issue to test on his end?

jsandai:
Tested on a Galaxy S26 Ultra. This fixes it.

For a control I first confirmed the bug on the official release: with a long enough transcript I was able to replicate the issue. I can scroll back roughly 13 messages and then the view snaps to the bottom so it seems that transcript length is what gates this. On a short thread nothing goes wrong at all, which is probably why it's been easy to miss.

I then built an APK from this branch and installed it over the top. With a 22 message transcript I can scroll to any earlier message and stop there, with no jump to the bottom. Bringing the keyboard up or dismissing it no longer moves the view either, which was my main concern with the change, since the composer takes focus when the screen opens and that alone used to snap you back. Sending a message still pulls the view to the bottom, which is what I'd want: you see what you sent and then the reply.

Both halves of the reported behaviour are gone and the one jump that remains is the one that should be there.

I've put an APK up for the reporter to try on their end: https://github.com/jsandai/SparkyFitness/releases/tag/pr2284-scroll-test

It's built from this branch at f82cb96de with nothing changed, and the release tag points at that same commit so it can be checked. It uses the dev application id, so it installs as a separate app alongside their existing one rather than replacing it, and it's signed with a test key. They'll need to sign in to their server on it, but the chat history lives server side so their existing transcript will be there, which matters because the length is what triggers the bug. arm64 only, to keep the download reasonable.

CodeWithCJ (approved):

PR #2289, opened 2026-08-28
mobile: allow plain HTTP to private/LAN/VPN hosts in production

### LINKED ISSUE #2288: [Feature]: Add Option to enable http server connection on mobile

Yes, I'd like to submit a PR!

Most people like me host sparkyfitness in homelab, while accessing it via tailscale / zerotier.

Currently only https:// is allowed not http://

Access http:// on mobile app.

Mobile app (iOS)

Blocking me from using SparkyFitness

Put a TLS-terminating reverse proxy in front of the homelab server so the mobile app always talks HTTPS, even over Tailscale/ZeroTier — e.g. Caddy/Nginx with a self-signed or ACME cert, or Tailscale's own built-in HTTPS certificates (`tailscale cert` / Tailscale Serve), which issue a real cert for your tailnet hostname and work seamlessly on iOS/Android.

### Additional context

Investigated in [PR #2289](https://github.com/CodeWithCJ/SparkyFitness/pull/2289): a private/local-network HTTP allowlist already existed in the app (`isPrivateOrLocalHost` in `serverUrl.ts`) but was dev-build-only and missing Tailscale's carrier-grade-NAT IP range (`100.64.0.0/10`). A fix extending that allowlist to production was proposed and implemented, but was **declined by the maintainer**: per [this comment](https://github.com/CodeWithCJ/SparkyFitness/issues/2288#issuecomment-5451751919), plain HTTP cannot be allowed in production regardless of network range — this is a hard requirement from Apple/Google for apps that access health data, camera, and passkeys, not something that can be relaxed per-host. The PR was closed with the analysis kept for reference.

**Resolution:** won't-fix. Self-hosters on Tailscale/ZeroTier should terminate TLS in front of their server (see Alternatives above) rather than expecting the app to accept plain HTTP.

Rahulsharma0810:
> **Status: Declined by maintainer.** Per [@CodeWithCJ's comment on #2288](https://github.com/CodeWithCJ/SparkyFitness/issues/2288#issuecomment-5451751919): plain HTTP cannot be allowed in production, full stop — this is a hard requirement from Apple/Google for apps that access health data, camera, and passkeys, not a policy decision that can be relaxed per-host or per-network-range. That requirement applies regardless of whether the destination is a private LAN, a Tailscale/ZeroTier address, or a public host, so the private-network carve-out this PR builds on is not an acceptable direction. Leaving this PR open for reference; not expecting it to be merged. For self-hosters on Tailscale/ZeroTier, the supported path is to put a TLS-terminating reverse proxy (e.g. Caddy, or Tailscale's own HTTPS certs) in front of the server instead.

---

Closes #2288

## Problem

Self-hosters who access their SparkyFitness server over Tailscale or ZeroTier (rather than a public HTTPS domain) couldn't use `http://` server URLs from the mobile app. The app already had a private/local-network allowlist (`isPrivateOrLocalHost` in `src/utils/serverUrl.ts`), but it was:

1. Gated behind `__DEV__`, so it only applied to development builds, never production installs.
2. Missing the carrier-grade NAT range (`100.64.0.0/10`), which is exactly the range Tailscale assigns tailnet addresses from — so even in dev builds, a Tailscale IP was treated as a public host and rejected.

Several call sites across the app (`apiClient`, `aiSettingsApi`, `authService`, `externalFoodSearchApi`, `healthDataApi`, `imageUploadClient`, `ReauthModal`, `ServerSettingsScreen`) also had their own ad-hoc `!__DEV__ && url.startsWith('http://')` checks instead of using the shared helper, so they needed the same fix in multiple places.

## Change

- `src/utils/serverUrl.ts`: added `carrierGradeNat` to `PRIVATE_IP_RANGES`, and removed the `__DEV__` gate from `getInsecureUrlError` so the private-network allowance applies identically in dev and production. Public hostnames/IPs are still always rejected over plain HTTP, in every build.
- Replaced the ad-hoc per-file `__DEV__` checks with the shared `isInsecureUrlBlocked` / `getInsecureUrlError` helpers, preserving each site's existing error type and message.
- Android (`plugins/withNetworkSecurityConfig.ts`): permit cleartext traffic at the OS layer. Android's `network-security-config` domain-config can only match exact hostnames/IPs, not CIDR ranges, so it can't express "private only" — that restriction is enforced entirely in the JS layer described above.
- iOS (`app.config.ts`): added `NSAllowsLocalNetworking: true` to `NSAppTransportSecurity`, keeping `NSAllowsArbitraryLoads: false`. This lets plain HTTP reach RFC 1918 and link-local LAN addresses without disabling ATS for arbitrary internet hosts.
- Updated `AGENTS.md` and `__tests__/utils/serverUrl.test.ts` to match.

## Security considerations

- The JS-layer allowlist (`isPrivateOrLocalHost` / `isInsecureUrlBlocked` in `src/utils/serverUrl.ts`) is now the actual security boundary on both platforms, since neither Android's nor iOS's OS-level transport-security config can express CIDR ranges. Public-domain plain HTTP is still rejected everywhere, in all builds — only the private-network allowance moved from dev-only to always-on. Every raw-fetch call site that carries auth headers or user data now routes through this shared guard (`apiClient`, `aiSettingsApi`, `authService`, `externalFoodSearchApi`, `healthDataApi`, `imageUploadClient`, `exerciseApi`, `ChatScreen`'s stream transport, `ReauthModal`, `ServerSettingsScreen`).
- On Android, the OS layer now permits cleartext traffic broadly (any `http://`, not just private ranges); the JS check is the sole enforcement point. This mirrors how the app already worked in dev builds.
- On iOS, `NSAllowsLocalNetworking` exempts **any IP-literal connection** (not scoped to RFC 1918/private ranges), per [Apple's docs](https://developer.apple.com/documentation/bundleresources/information-property-list/nsapptransportsecurity/nsallowslocalnetworking), plus unqualified and `.local` hostnames. That means a raw Tailscale/CGNAT IP literal (e.g. `100.64.x.x`) should work through this flag alone — an earlier version of this PR incorrectly assumed CGNAT needed `NSAllowsArbitraryLoads: true`; that's been corrected. The one remaining iOS gap is a *qualified* mesh-VPN hostname (e.g. a Tailscale MagicDNS name like `device.tailnet.ts.net`), which would still need HTTPS or an `NSExceptionDomains` entry. This is based on Apple's documentation and has not yet been verified on a physical device.
- Passkey/WebAuthn sign-in requires a true secure context, which plain HTTP never satisfies even for a private IP. `loginWithPasskey` and `ReauthModal`'s passkey flow use a dedicated stricter check (`isPasskeyUrlBlocked`, HTTPS-only) rather than the general private-network allowance, so users get an upfront error instead of a browser session that silently fails after opening.
- The `.local`/`.lan`/`.internal`/`.home.arpa` hostname allowlist in `isPrivateOrLocalHost` is a string match, not a DNS-resolved check, so it can't detect DNS rebinding (a malicious record pointing such a hostname at a public IP). This is an accepted residual risk: the hostname is the user's own server address, typed once into their own trusted Settings config — not attacker-supplied per request like a CORS `Origin` header — so exploiting it requires social-engineering a user into entering a malicious hostname as their own server.

## Testing

- `pnpm exec jest --watchman=false --runInBand` (SparkyFitnessMobile) — all suites pass, including new/updated cases in `serverUrl.test.ts` covering the CGNAT range boundaries and the dev/production parity of `getInsecureUrlError`.
- `pnpm run validate` (typecheck, lint, i18n audit, native locales) — passes.

## Out of scope

`SparkyFitnessServer/utils/corsHelper.ts` has a very similar `isPrivateNetworkAddress` allowlist (used for CORS), which doesn't include `carrierGradeNat` either. It isn't on the mobile app's request path, so it's unaffected by this bug, but it could use the same one-line addition for consistency if maintainers want that as a follow-up.

---
## ✅ Required Checklist (restored automatically)

> has been added back below. **Please tick each box and do not delete this section** —
> no new commit is needed.

- [ ] **[MANDATORY for UI changes] Screenshots**: I have attached Before/After screenshots below. (No UI changes — this PR is a transport/security-layer fix; there is nothing to screenshot.)
- [ ] **[MANDATORY for Mobile changes] Tested on device or emulator**: I have verified the changes work on iOS or Android. (Verified via unit tests and `pnpm run validate`; not yet manually verified on a physical device against a real Tailscale/ZeroTier connection.)

## Summary by CodeRabbit

* **New Features**
  * Self-hosted servers on private LAN, local, and mesh VPN networks can use HTTP connections in all environments.
  * Added support for carrier-grade NAT addresses used by mesh VPNs, with Android support and continued HTTPS requirements on iOS.
  * iOS and Android now support permitted local HTTP connections.

* **Bug Fixes**
  * Public HTTP connections are consistently blocked across server setup, authentication, chat, synchronization, uploads, and API requests.
  * Passkey sign-in now always requires HTTPS.
  * Updated server settings and chat messaging to clearly explain HTTPS requirements.


Rahulsharma0810:
Thanks for the review, @coderabbitai — addressed all 5 findings in 95ddb7fed:

1. **AGENTS.md wording** — clarified that the JS layer (`src/utils/serverUrl.ts`) is the real security boundary on both platforms, since neither OS transport config can express a 'private-only' exception (Android's cleartext flag is now open to any host; iOS's `NSAllowsLocalNetworking` exempts any IP literal, not just private ranges).

2. **Missing transport guards** — found and fixed 3 more raw-fetch call sites that bypassed the shared check entirely: `exerciseApi.ts`'s `createExercise`/`updateExercise` (multipart mutations that don't go through `apiFetch`), and `ChatScreen.tsx`'s chat-stream transport (previously had no transport check at all).

3. **Passkey flows need HTTPS-only, not just private-network** — added a new `isPasskeyUrlBlocked` helper (strict `https://` check) used only in `authService.loginWithPasskey` and `ReauthModal.handlePasskeySignIn`. WebAuthn requires a true browser secure context, which plain HTTP to a private IP does not satisfy — this avoids opening a browser session that then silently fails. `loginWithOidc` intentionally keeps the general `isInsecureUrlBlocked` check since OIDC has no secure-context requirement.

4. **DNS-rebinding on the `.lan`/`.local`/`.internal`/`.home.arpa` hostname allowlist** — documented as an accepted residual risk rather than fixed. React Native has no reliable synchronous DNS-resolution API without a native module, and unlike a CORS `Origin` header, this hostname isn't attacker-supplied per request — it's the user's own server address, typed once into their own trusted Settings config. Exploiting it requires social-engineering a user into entering a specific malicious hostname as their own server, not passive exposure.

5. **iOS `NSAllowsLocalNetworking` / IP-literal ATS behavior** — this one changed my understanding, thanks for pushing on it. Per [Apple's docs](https://developer.apple.com/documentation/bundleresources/information-property-list/nsapptransportsecurity/nsallowslocalnetworking), `NSAllowsLocalNetworking` exempts **any** IP-literal connection (not scoped to RFC1918/private ranges), plus unqualified and `.local` hostnames. So a raw Tailscale IP (`100.64.x.x`) should actually work on iOS via this flag alone — my earlier assumption that CGNAT needed `NSAllowsArbitraryLoads` was wrong and has been corrected in `app.config.ts` and `AGENTS.md`. The one real remaining iOS gap is a *qualified* mesh-VPN hostname (e.g. a Tailscale MagicDNS name like `device.tailnet.ts.net`, which has multiple labels) — that would still need HTTPS or an `NSExceptionDomains` entry. Note this is based on Apple's documentation, not yet verified on a physical device.

All existing tests plus the affected suites (`serverUrl`, `services/api`, `ChatScreen`, `ReauthModal`) and `pnpm run validate` (typecheck/lint/i18n audit) pass.

Rahulsharma0810:
Pushed b7e4c21ec addressing the remaining findings:

- **ChatScreen.tsx** — a blocked (insecure) server URL is now tracked separately from "no server configured", so the chat screen shows an accurate HTTPS-required message instead of the generic "no server" prompt when a configured server is rejected by the transport guard.
- **iOS + CGNAT (Tailscale/ZeroTier)** — after digging further into Apple's ATS docs and your own research turning up conflicting guidance, I've made the JS layer conservative here: carrier-grade NAT (100.64.0.0/10) is allowed over HTTP on Android (where the OS-level cleartext permission isn't range-scoped anyway), but on iOS it still requires HTTPS until real-device testing confirms `NSAllowsLocalNetworking`'s actual behavior for that range. This is enforced via a new `IOS_UNVERIFIED_PRIVATE_RANGES` set in `src/utils/serverUrl.ts`, with matching test coverage (`Platform.OS` mocked per-case) and updated docs in `app.config.ts`/`AGENTS.md`.

DNS-rebinding on the `.local`/`.lan`/`.internal`/`.home.arpa` hostname allowlist remains a documented, accepted residual risk (no reliable synchronous DNS resolution in RN without a native module; the threat model requires the user to type a malicious hostname into their own trusted server config, not passive exposure to attacker-controlled input).

Rahulsharma0810:
Closing per maintainer feedback on #2288 — plain HTTP isn't an option for this app regardless of network range (Apple/Google requirements for health data, camera, and passkeys). Leaving the description in place as a record of the analysis for anyone who hits this same request in the future.

PR #2334, opened 2026-08-31
fix(food-search): paginate local food search so later matches stay reachable

### LINKED ISSUE #2333: [Bug]: Add-to-meal food search is capped at 10 results with no pagination

Yes, I'd like to submit a PR!

### What happened?

A food that exists in the database can't be found from the food search used to add a food to a meal, even though searching the exact same term in the Food library finds it.

The local "Your Foods" section of that search is capped at 10 rows, with no pagination and no Load more, so anything that sorts 11th or later is unreachable. Results are ordered by name, not by relevance or recency, so a newly added food lands wherever its name falls alphabetically. The Food library runs the same query but paginates it, so the food shows up there on page 2. The mobile app requests 20 per page with pagination, so it shows up there too. Only the web add-to-meal search hides it.

The cap can't be raised, because the preference that's supposed to control it is dead end to end. `foodCoreService.searchFoods` reads `userPreferences?.food_display_limit`, but that column was renamed to `item_display_limit` by migration `20250720201800_rename_food_display_limit_to_item_display_limit.sql` and the read was never updated. It's always `undefined`, so the limit falls through to whatever the client sent. The client sends `foodDisplayLimit` from `PreferencesContext`, which is set from `data.food_display_limit || 10`, so that's always 10 as well, and there's no control for it in Settings. The preference is declared, defaulted, sent on save and referenced in the service, and has no effect anywhere.

### Steps to reproduce

1. Create at least 11 foods whose brand plus name contain a common word, for example "bread".
2. Create one more food whose name sorts alphabetically last among that set, for example brand "Test Bakery", name "Zzz White Bread".
3. Go to Foods (the food library) and search "bread". The new food is there, on page 2 at the default 10 per page.
4. Go to the diary, pick a meal, and open the add food search. Search "bread".
5. Under "Your Foods" exactly 10 results are listed, the new food isn't one of them, and there's no pagination or Load more to reach it.

Optional, for the related filter problem: set the ownership dropdown in that same search to "Mine". The filter runs in the browser against the 10 rows the server already truncated, so the list can collapse to a couple of results even when you own dozens of matches.

### Expected behavior

Every food matching the search term should be reachable from the add-to-meal search, the same way it already is in the Food library and in the mobile app. The results should paginate (a Load more, matching what the online provider results already do), and the ownership filter should be applied by the server before the limit rather than in the browser after it.

Docker Compose

### SparkyFitness version

v1.6.3

### Which area is affected?

Web frontend, Server / API

```yaml

```

```shell

```

_No response_

### Anything else?

Code references, all on `main` at 82cfee9:

- `SparkyFitnessFrontend/src/components/FoodSearch/FoodSearch.tsx` calls `useDatabaseFoodSearchQuery(debouncedSearchTerm, foodDisplayLimit, ...)`, which hits `GET /foods?name=...&broadMatch=true&limit=10`.
- `SparkyFitnessServer/services/foodCoreService.ts` line 89: `const limit = userPreferences?.food_display_limit || limitFromRequest;`
- `SparkyFitnessServer/db/migrations/20250720201800_rename_food_display_limit_to_item_display_limit.sql` is what renamed the column out from under that read.
- `SparkyFitnessServer/models/food.ts` `searchFoods` orders by exact substring first, then `f.name ASC`, then applies `LIMIT`. It takes no ownership filter and returns no total count.
- `SparkyFitnessServer/models/food.ts` `getFoodsWithPagination` runs the same WHERE and the same ORDER BY, but adds `foodFilter`, `LIMIT`/`OFFSET` and a stable `f.id ASC` tiebreak, and returns `totalCount`. The Food library and `SparkyFitnessMobile/src/services/api/foodsApi.ts` already use it.
- `SparkyFitnessFrontend/src/components/FoodSearch/FoodSearch.tsx` applies the ownership filter client side via `filterItems(searchFoodsFavFirst, ownershipFilter, user?.id)`, after the server has already truncated.

Related: the same Load more affordance for single-provider online results was added in #1704. This is the local-results equivalent.

jsandai:
## Description

The local food search in the add-food dialog was capped at a fixed number of rows with no pagination, so a match sorting past that cap couldn't be reached at all, even though the Food library listed it and the mobile app found it. The ownership filter made it worse by filtering in the browser, after the server had already truncated.

The dialog's local search now uses `/foods/foods-paginated`, the endpoint the Food library and the mobile app already use, with `item_display_limit` as the page size and a "Load more" that mirrors the existing online-provider results. The ownership filter goes with the request, so it narrows the match set before the page boundary instead of after it.

Separately, the search limit on `GET /foods` and `/foods/search` read `userPreferences.food_display_limit`. Migration `20250720201800_rename_food_display_limit_to_item_display_limit.sql` renamed that column, so the read resolved to `undefined` and the preference never applied to anything. It now reads `item_display_limit`. The frontend mirror of the same dead preference is removed, and the online-provider page size reads the surviving preference instead.

Linked Issue: Closes #2333

## How to Test

1. Check out this branch, then run `pnpm install` and start the stack.
2. Create enough foods that a common search word matches more than `item_display_limit` of them (default 10), and make sure one of them sorts alphabetically last, for example brand "Test Bakery", name "Zzz White Keto Bread".
3. Open the diary, pick a meal, and click the add-food button. Search for the common word.
4. The local "Your Foods" list now ends with a "Load more" button. Click it and confirm the last-sorting food appears.
5. Set the ownership dropdown to "My Foods" and repeat. The count should reflect every food you own that matches, not just the ones that survived the first page.

Automated coverage:

- `SparkyFitnessFrontend/src/tests/hooks/useDatabaseFoodSearchQuery.test.tsx` covers reaching a match past the first page, stopping at `totalCount`, and sending the ownership filter to the server.
- `SparkyFitnessServer/tests/foodCoreService.test.ts` covers the preference being honoured and the fallback when none is set.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

<details>

2026-08-31 16_33_24-Greenshot

</details>

## Notes for Reviewers

No new endpoints or tables, so the Zod and RLS items don't apply. No translation keys were added: the button reuses the existing `enhancedFoodSearch.loadMore`.

Two things:

Removing the dead `food_display_limit` preference leaves `item_display_limit` governing both the landing sections and the search page size. Defaults are unchanged at 10, so this is not a behaviour change out of the box, but it does mean the existing Settings control now affects search results as well. Happy to split that into its own change if we'd rather keep the preferences separate.

`MealBuilder` used the old capped search helper for a different job: matching an incoming meal food against an existing food by name before creating a duplicate. That call is preserved as `foodNameLookupOptions` with its own cache key rather than folded into the paginated search. Note that this lookup reads the response as an array while the endpoint returns `{ searchResults: [...] }`, so the match never succeeds. That predates this PR and isn't changed here, but it looks like a real bug worth its own issue.

The search ordering is still exact-substring-first then name, with no recency or relevance weighting. That's unchanged here, and it's the reason a newly added food can land on a later page in the first place.

## Summary by CodeRabbit

* **New Features**
  * Added paginated food search with server-side ownership filtering.
  * Added a “Load more” option for the “Your Foods” section.
  * Improved name matching when adding imported or unsaved meal ingredients.

* **Bug Fixes**
  * Food searches now honor the configured item display limit.
  * Search results refresh correctly when the display limit changes.

* **Improvements**
  * Renamed the food display limit preference to the broader item display limit.
  * Improved consistency across provider searches and cached results.

### DIFF
--- a/SparkyFitnessFrontend/src/api/Foods/foodService.ts
+++ b/SparkyFitnessFrontend/src/api/Foods/foodService.ts
@@ -141,17 +141,17 @@ export const getRecentAndTopFoods = async (
   return apiCall(`/foods?${params.toString()}`);
 };
 
-export const searchDatabaseFoods = async (
-  term: string,
-  limit: number,
-  mealType?: string
-) => {
+/**
+ * Name lookup used to match an incoming meal food against an existing food
+ * before creating a duplicate. Not the dialog's user-facing search: that one
+ * paginates through `loadFoods`.
+ */
+export const lookupFoodsByName = async (term: string, limit: number) => {
   const params = new URLSearchParams({
     name: term,
     broadMatch: 'true',
     limit: limit.toString(),
   });
-  if (mealType) params.append('mealType', mealType);
 
   return apiCall(`/foods?${params.toString()}`);
 };
--- a/SparkyFitnessFrontend/src/api/keys/meals.ts
+++ b/SparkyFitnessFrontend/src/api/keys/meals.ts
@@ -24,8 +24,12 @@ export const foodKeys = {
     [...foodKeys.lists(), { searchTerm, filter, page, limit, sort }] as const,
   recentTop: (limit: number, mealType?: string) =>
     [...foodKeys.all, 'recentTop', limit, mealType] as const,
-  databaseSearch: (term: string, limit: number, mealType?: string) =>
-    [...foodKeys.all, 'search', term, limit, mealType] as const,
+  databaseSearch: (term: string, pageSize: number, filter: MealFilter) =>
+    [...foodKeys.all, 'search', term, pageSize, filter] as const,
+  // Separate namespace from `databaseSearch`: that one is an infinite query
+  // whose cached shape is {pages}, this one resolves to a single response.
+  nameLookup: (term: string, limit: number) =>
+    [...foodKeys.all, 'nameLookup', term, limit] as const,
   favorites: () => [...foodKeys.all, 'favorites'] as const,
 };
 
--- a/SparkyFitnessFrontend/src/components/FoodSearch/FoodSearch.tsx
+++ b/SparkyFitnessFrontend/src/components/FoodSearch/FoodSearch.tsx
@@ -172,7 +172,6 @@ const EnhancedFoodSearch = ({
     foodSearchAllProvidersDefault,
     defaultBarcodeProviderId,
     itemDisplayLimit,
-    foodDisplayLimit,
     nutrientDisplayPreferences,
     energyUnit,
     convertEnergy,
@@ -260,13 +259,22 @@ const EnhancedFoodSearch = ({
     topMeals,
     isLoading: isLoadingRecentMeals,
   } = useRecentAndTopMealsQuery(itemDisplayLimit, showMeals && isSearchEmpty);
-  const { data: searchData, isFetching: isFetchingSearch } =
-    useDatabaseFoodSearchQuery(
-      debouncedSearchTerm,
-      foodDisplayLimit,
-      mealType,
-      showLocalFoods && !!debouncedSearchTerm.trim()
-    );
+  // Local food search is paginated on the server, page size item_display_limit.
+  // The ownership filter goes with the request so it narrows the match set
+  // before the page boundary; filtering the page in the browser would hide
+  // matches that simply sat on a later page.
+  const {
+    data: searchData,
+    isFetching: isFetchingSearch,
+    hasNextPage: hasMoreLocalFoods,
+    fetchNextPage: fetchMoreLocalFoods,
+    isFetchingNextPage: isLoadingMoreLocalFoods,
+  } = useDatabaseFoodSearchQuery(
+    debouncedSearchTerm,
+    itemDisplayLimit,
+    ownershipFilter,
+    showLocalFoods && !!debouncedSearchTerm.trim()
+  );
 
   // Starred foods and meals. Shared cache with the row star in FoodResultCard,
   // so this is one fetch, not two.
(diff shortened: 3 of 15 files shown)


jsandai:
@coderabbitai

**`itemDisplayLimit` missing from `allProvidersFoodSearchKey`**: fixed. You're right that it's the `pageSize` for the `PAGE_SIZE_PROVIDERS`, so omitting it from the key served results fetched under the previous limit. The omission predates this PR, but it was harmless while `foodDisplayLimit` was a dead constant pinned at 10. Making the preference live is what turns it into a staleness bug, so it belongs here. Added the limit to the key plus `src/tests/hooks/useAllProvidersFoodSearch.test.tsx`, which covers both directions: changing the limit refetches with the new page size, and leaving it alone still serves the cache. Removing the argument again fails the first of those by name.

**`test-utils.tsx` for the provider wrapper**: I'd rather leave this one, because the injection path it assumes doesn't exist. `test-utils.tsx` exports only `renderWithClient(ui, options)`, which is a `render` helper for components and builds its query client internally, with no way to pass one in. These are `renderHook` tests, so there's nothing there to use. The existing convention agrees: `useBackups`, `useMealsFavoritesInvalidation` and `useOnboarding` all build a local `QueryClientProvider` wrapper, and no hook test imports `test-utils`. Happy to revisit if you'd like a `renderHook` helper added to the shared utils, but that feels like its own change rather than part of this fix.

Full frontend `pnpm run validate` passes, 1028 tests across 107 suites, and the server suite is green at 3781.

@coderabbitai review

jsandai:
`expect.arrayContaining([10])` only proved the number appeared somewhere in the call, and `searchFoodsV2` takes `page` immediately before `pageSize`, so a limit passed into the wrong slot would still have satisfied it. Both assertions now pin the whole call with `toHaveBeenNthCalledWith`. To confirm the tightening is real rather than cosmetic, I moved `pageSize` into the `page` slot: the old check passes on that mutation, the new one fails `refetches with the new page size when the display limit changes` by name.

Frontend `pnpm run validate` passes, 1028 tests across 107 suites.

Nothing to do on the `test-utils.tsx` comment, since CR withdrew it in the meantime. That review was submitted about a minute before the withdrawal, so it just crossed in flight.

@coderabbitai review

CodeWithCJ (approved):

Issue #2078, opened 2026-08-10
[Bug]: Barcode lookup resolves differently on web vs mobile when no default provider is set; FatSecret offered despite Premier-only barcode endpoint


brad-thomson:
### Summary

The same barcode resolves differently depending on whether it's scanned from the web UI or the mobile app, because the two clients disagree about what to send when the user hasn't chosen a barcode provider. Separately, FatSecret is offered as a barcode provider even though its barcode endpoint is Premier-only, and its failure masks the fallback result.

Version: v1.6.1.

### 1. Web auto-selects a provider; mobile sends none

`SparkyFitnessFrontend/src/components/FoodSearch/FoodSearch.tsx:448-454`

```ts
const selectedBarcodeProvider =
  barcodeProviderId ||
  defaultBarcodeProviderId ||
  foodDataProviders.find((p) =>
    ['openfoodfacts', 'usda', 'fatsecret', 'yazio'].includes(p.provider_type)
  )?.id ||
  null;
```

With no explicit selection and no `default_barcode_provider_id` set, the web falls through to *the first barcode-capable provider in the list* and sends it as `?providerId=`. The mobile app sends no `providerId` at all.

Server-side (`SparkyFitnessServer/services/foodCoreService.ts:862-871`), `resolvedProviderId = providerId || userPreferences?.default_barcode_provider_id`. When nothing resolves, `provider` stays null, every primary branch is skipped, and the lookup goes straight to the OpenFoodFacts fallback.

So the two clients take different paths for identical input. Observed on my instance with `default_barcode_provider_id` unset:

| Barcode | Client | Provider sent | Result |
|---|---|---|---|
| `8801007957869` | web | USDA (auto-picked) | no product |
| `8801007957869` | mobile | *(none)* | found — "Vegetable Dumplings Japchae" |
| `8801392036552` | web | auto-picked | no product |
| `8801392036552` | mobile | *(none)* | found — "Rice Balls Bibimbap Style" |

Both products are in OpenFoodFacts (verified directly against `world.openfoodfacts.org/api/v2/product/…`). The web scans missed because they were routed to USDA, which is US-only and has no entry for these EANs.

Setting a `Default Barcode Provider` explicitly works around it, but the unset default shouldn't produce client-dependent behavior.

**Suggested fix:** when the user hasn't explicitly chosen a provider, have the web send no `providerId` and let the server resolve it — matching mobile and keeping the resolution logic in one place.

[...]

PR #2347, opened 2026-09-02
fix(mobile): stop the food-search provider overriding the barcode provider

### LINKED ISSUE #2339: [Bug]: Barcode scanning of this item fails on app, succeeds on web

_No response_

### What happened?

I tried scanning this protein powder on the app about a dozen times and it never saved to food.
It worked on the web portal with the first try.
App Food settings: 
Default food source: "First Available"
Barcode Scanning: USDA

Image

### Steps to reproduce

Scan the barcode with the app

### Expected behavior

To add the item to the Food list

Docker (manual)

### SparkyFitness version

1.6.4

### Which area is affected?

Mobile app (Android)

```yaml
services:
  sparkyfitness-db:
    image: postgres:15-alpine
    container_name: SparkyFitness-DB
    hostname: sparkyfitness-db
    security_opt:
      - no-new-privileges:true
    healthcheck:
      test: ["CMD", "pg_isready", "-q", "-d", "sparky", "-U", "sparkyuser"]
      timeout: 45s
      interval: 10s
      retries: 10
    environment:
      POSTGRES_DB: sparky
      POSTGRES_USER: sparkyuser
      POSTGRES_PASSWORD: sparkypass
    volumes:
      - /volume1/docker/sparkyfitnessdb:/var/lib/postgresql/data:rw
    restart: on-failure:5

  sparkyfitness-web:
    image: codewithcj/sparkyfitness:latest
    container_name: SparkyFitness-WEB
    healthcheck:
      test: ["CMD-SHELL", "nc -z 127.0.0.1 80 || exit 1"]
      interval: 10s
      timeout: 5s
      retries: 3
      start_period: 90s
    ports:
      - 3006:80
    depends_on:
      - sparkyfitness-server
    restart: on-failure:5

  sparkyfitness-server:
    image: codewithcj/sparkyfitness_server:latest
    container_name: SparkyFitness-SERVER
    environment:
      SPARKY_FITNESS_LOG_LEVEL: INFO
      SPARKY_FITNESS_APP_DB_USER: sparkyuser
      SPARKY_FITNESS_APP_DB_PASSWORD: sparkypass
      SPARKY_FITNESS_DB_USER: sparkyuser
      SPARKY_FITNESS_DB_PASSWORD: sparkypass
      SPARKY_FITNESS_DB_HOST: sparkyfitness-db
      SPARKY_FITNESS_DB_NAME: sparky
      SPARKY_FITNESS_DB_PORT: 5432
      SPARKY_FITNESS_SERVER_PORT: 3010
      SPARKY_FITNESS_EXTRA_TRUSTED_ORIGINS: https://[redacted]
      ALLOW_PRIVATE_NETWORK_CORS: true
      NODE_ENV: production
      TZ: Europe/Bucharest
      SPARKY_FITNESS_API_ENCRYPTION_KEY: [redacted]
      BETTER_AUTH_SECRET: [redacted]
      SPARKY_FITNESS_FRONTEND_URL: https://[redacted]
      SPARKY_FITNESS_DISABLE_SIGNUP: false #or true to Disable Sign Up after first registration.
    ports:
      - 4059:3010
    depends_on:
      - sparkyfitness-db
    restart: on-failure:5
```

```shell

```

_No response_

### Anything else?

_No response_

jsandai:
## Description

On the mobile app, a barcode scan started from the food search screen ignores the Barcode Scanning provider setting, so the same barcode can resolve through a different provider than the one the user picked.

`FoodSearchScreen` forwarded the currently selected food-search provider as `providerId` when navigating to `FoodScan`. The server resolves the barcode provider as `providerId || default_barcode_provider_id`, so an explicit param always wins and the Barcode Scanning preference never applies on that path. The "+" -> Scan Food entry point sends no `providerId` and already behaves correctly, so the same barcode resolved differently depending on which entry point was used. With Default Food Source set to "First Available" the screen falls back to `providers[0]`, which pins barcode lookups to whichever provider happens to sort first.

This drops the param so both entry points let the server resolve the preference, matching the web client, which prefers `defaultBarcodeProviderId`.

Linked Issue: Closes #2339

## How to Test

1. In app Settings, set Default Food Source to a provider that isn't your barcode provider (or to "First Available"), and set Barcode Scanning to a different provider, for example USDA.
2. Open a meal, tap "+" -> Scan Food, and scan a barcode. Note which provider the result comes from. This path was already correct.
3. Go back, open Food Search, and tap the scan icon in the search bar. Scan the same barcode.
4. Before this change, step 3 used the food-search provider and could fail or return a different item than step 2. After it, both paths use the Barcode Scanning provider and return the same result.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

No visual change. The fix only affects which provider the scanner asks, so there's nothing to show side by side.

## Notes for Reviewers

The removed branch also handled the All Providers sentinel, which is why the old code read as deliberate. That guard is no longer needed: with no `providerId` at all, the sentinel can never reach the scanner, and the server's own default resolution takes over.

Test coverage: `__tests__/screens/FoodSearchScreen.test.tsx` gets a case asserting the navigate payload has no `providerId`. It carries a positive control that the screen really has settled on a food-search provider first, so the assertion can't pass just because nothing was selected yet.

**A follow-up worth deciding separately, not in this PR.** After this change nothing in the app sets the `providerId` route param on `FoodScan`, but `types/navigation.ts` still declares it and `FoodScanScreen` still reads it, so it's now a consumer with no producer. `EditBarcodeScreen` already calls `lookupBarcodeV2(barcode)` with no provider, so the one-arg form is the norm in mobile and this screen was the odd one out.

I've left the param in place deliberately, to keep this a tight bugfix and because the web `BarcodeScannerDialog` has a per-scan provider dropdown that mobile doesn't yet, so the slot has a plausible future consumer if mobile ever reaches parity. The argument the other way is that an unset `providerId` slot on a barcode screen is what invited this bug: someone saw an empty parameter and wired the nearest plausible value into it, and leaving it keeps that available to the next person. Stripping it would spread the diff into `FoodScanScreen` and the navigation types, which turns a bugfix into a fix plus refactor.

Happy to strip it in a follow-up if we'd prefer the preference to be the single source of truth for provider selection.

## Summary by CodeRabbit

- **Bug Fixes**
  - Barcode scanning now uses the server-configured provider instead of inheriting the provider selected for food search.
  - Scanning works consistently when a specific provider or “All Providers” is selected.

### DIFF
--- a/SparkyFitnessMobile/__tests__/screens/FoodSearchScreen.test.tsx
+++ b/SparkyFitnessMobile/__tests__/screens/FoodSearchScreen.test.tsx
@@ -861,6 +861,36 @@ describe('FoodSearchScreen', () => {
     );
   });
 
+  it('does not forward the food-search provider to the barcode scanner', () => {
+    // A provider is actively selected in food search (no saved default, so the
+    // screen falls back to providers[0] — the "First Available" case). The
+    // barcode scanner must not inherit it: the server has to be left free to
+    // resolve default_barcode_provider_id, or an explicit Barcode Scanning
+    // preference gets silently overridden by Default Food Source.
+    mockUseExternalProviders.mockReturnValue(fatSecretProvider);
+    const screen = render(
+      <SafeAreaProvider initialMetrics={{ insets, frame }}>
+        <FoodSearchScreen navigation={navigation} route={route} />
+      </SafeAreaProvider>
+    );
+    fireEvent.press(screen.getByLabelText('Scan Food'));
+
+    // Positive control: the screen really has settled on FatSecret for online
+    // search. Without this the absence check below could pass simply because no
+    // provider was selected yet.
+    expect(mockUseExternalFoodSearch).toHaveBeenLastCalledWith(
+      expect.anything(),
+      'fatsecret',
+      expect.objectContaining({ providerId: 'p1' })
+    );
+
+    const scanCall = navigation.navigate.mock.calls.find(
+      ([screenName]: [string]) => screenName === 'FoodScan'
+    );
+    expect(scanCall).toBeDefined();
+    expect(scanCall[1]).not.toHaveProperty('providerId');
+  });
+
   describe('persisted "All Providers" default', () => {
     const twoProviders = {
       providers: [
--- a/SparkyFitnessMobile/src/screens/FoodSearchScreen.tsx
+++ b/SparkyFitnessMobile/src/screens/FoodSearchScreen.tsx
@@ -385,21 +385,13 @@ const FoodSearchScreen: React.FC<FoodSearchScreenProps> = ({
       // Preserve the originating meal type (MealTypeDetail → FoodSearch → scan).
       mealTypeId: mealTypeId ?? undefined,
       mealPlanTarget,
-      // Never forward the All Providers sentinel as a real provider; the scanner
-      // should fall back to its default provider in that mode.
-      providerId:
-        selectedProvider === ALL_PROVIDERS_VALUE
-          ? undefined
-          : (selectedProvider ?? undefined),
+      // Deliberately no providerId. The food-search provider is not the barcode
+      // provider, and the server treats an explicit providerId as winning over
+      // default_barcode_provider_id, so forwarding it here silently overrode the
+      // user's Barcode Scanning setting. Let the server resolve the preference,
+      // matching the "+" → Scan Food entry point.
     });
-  }, [
-    navigation,
-    date,
-    mealPlanTarget,
-    mealTypeId,
-    selectedProvider,
-    selectionPickerMode,
-  ]);
+  }, [navigation, date, mealPlanTarget, mealTypeId, selectionPickerMode]);
 
   // Only the custom-header path opens the JS menu; on the native path the
   // system presents a UIMenu from the header item directly.


CodeWithCJ:
what happens when the default provider for barcode doesn't result in any food? Will it auto try or we have option to change the provider temp on the fly?

jsandai:
> what happens when the default provider for barcode doesn't result in any food? Will it auto try or we have option to change the provider temp on the fly?

**Auto-retry:** there's one fallback, not a general one. `lookupBarcode` tries the resolved barcode provider, and if that returns no match it falls back to Open Food Facts, gated on the user's "fallback to Open Food Facts" preference (`foodCoreService.ts:1081`). It doesn't try the user's other configured providers. So with Barcode Scanning set to USDA and FatSecret also configured, a USDA miss goes to OFF and never to FatSecret.

One caveat that's easy to miss: if any provider throws an error carrying a status, the first such error is captured and rethrown at the end, even when the OFF fallback already ran and legitimately found nothing (`foodCoreService.ts:1136`). That looks deliberate, the comment reads "report a misconfiguration rather than not found", but it does mean a broken primary provider masks a clean miss. That's the mechanism behind #2078: FatSecret's barcode endpoint is Premier tier, so a free key throws rather than returning empty, and the user sees an error instead of "not found".

**Changing provider on the fly:** web has it, mobile doesn't. `BarcodeScannerDialog` has the per-scan provider dropdown. On mobile there's no picker on the scan screen. The `providerId` route param on `FoodScan` is what a picker would feed, and `FoodScanScreen` still reads it, but after this PR nothing sets it. That's the follow-up in the description: either mobile gets the same picker and the param gets a real producer, or the param goes away. Happy to do either, whichever you prefer.

This PR doesn't change any of that. It only stops the food-search provider being sent as the barcode provider, so the Barcode Scanning preference applies on both entry points the way it already did on "+ > Scan Food".

CodeWithCJ:
Merging this for now. We can park the drop down for another PR.

CodeWithCJ (approved):

PR #2360, opened 2026-09-03
fix(exercise): data sync orphaned child writes

### LINKED ISSUE #2351: [Bug]: Health Connect workout sync fails with RLS 42501 on exercise_entry_sets / exercise_entry_activity_details — parent entry deleted mid-request by non-atomic sync pre-cleanup

No, but I can help test.

### What happened?

Android Health Connect workout sync fails on the server with PostgreSQL error `42501` (`new row violates row-level security policy`) on two tables — `exercise_entry_sets` and `exercise_entry_activity_details`. The parent `exercise_entries` write never errors, so the sync reports partial success while sets and activity details are silently lost.

The error message is misleading: **this is not a permissions problem.** The affected user is a plain solo account writing their own data, and every RLS policy is correctly installed. The real cause is that the parent `exercise_entries` row *disappears* between the parent write and the child writes. Because both child tables resolve their parent through an `EXISTS` subquery, a missing parent evaluates as "not permitted" rather than "not found".

### Crux of the investigation

Condensed summary of what was explored and what it established.

**1. The version in the logs is not the version I thought I was running.** I initially reported this against `v1.6.1`. Matching all nine stack-trace frames against each release tag shows the line numbers match `v1.6.2`/`v1.6.3`/`v1.6.4` exactly (identical across those three for these files) and match neither `v1.6.1` nor `main`. Confirmed deployment is **v1.6.4**.

| Frame | v1.6.1 | v1.6.2–v1.6.4 |
|---|---|---|
| `exerciseEntry.ts:581` | `snapshot.modality,` | `await client.query(setsQuery);` ✅ |
| `exerciseEntry.ts:655` | `options` | `_updateExerciseEntryWithClient(` ✅ |
| `exerciseEntry.ts:786` | `client: any,` | `_createExerciseEntryWithClient(` ✅ |
| `healthDataHandlers.ts:1522` | *(past EOF)* | `createExerciseEntry(` ✅ |
| `healthDataHandlers.ts:1563` | *(past EOF)* | `createActivityDetail(ctx.userI…` ✅ |

**2. It is not an RLS/permissions defect.** Verified directly against the database:

- `SELECT ... FROM family_access WHERE family_user_id = '<user>'` → **0 rows**. With no delegation, `has_diary_access(owner)` short-circuits on `authenticated_user_id() = owner_uuid` and cannot return false for the user's own rows.
- `pg_policies` returns all 7 expected policies on `exercise_entries`, `exercise_entry_sets`, `exercise_entry_activity_details`. Nothing is missing or stale.
- `SELECT ... FROM exercise_entries WHERE id = '<id from the failing insert>'` → **0 rows**. The parent is genuinely absent.

**3. The error code actively hides the cause.** The log's `routine: 'ExecWithCheckOptions'` confirms the RLS `WITH CHECK` is evaluated inline, while the foreign key on `exercise_entry_id` is an `AFTER ROW` trigger that only fires at end of statement. So a vanished parent *always* reports as `42501` and never as the foreign-key violation that would have pointed straight at the truth. Considerable time was lost chasing a permissions bug that does not exist.

**4. The deleter is the sync's own pre-cleanup.** `processHealthData` step 0 calls `deleteExerciseEntriesByEntrySourceAndDate` (`measurementService.ts:231-239`), which range-deletes that source's activity details, sets, then entries across the batch's `[min,max]` day span.

**5. Nothing is atomic.** The pre-cleanup, the parent write, and the child writes are **three separate transactions on three separate pooled connections** (`deleteExerciseEntriesByEntrySourceAndDate`, `createExerciseEntry`, `createActivityDetail` each call `getClient()` and `BEGIN`/`COMMIT` independently). Under PostgreSQL's default `READ COMMITTED`, each statement takes a fresh snapshot — so a concurrent commit can remove the parent even *within* one `createExerciseEntry` transaction, between its existence check and its child insert.

**6. This failure mode is already known to the project.** The mobile client mitigates it client-side by packing workouts into disjoint contiguous day ranges, and says so explicitly (`SparkyFitnessMobile/src/services/api/healthDataApi.ts:157-160`, `:186-195`):

> Exercise/Workout use server-side delete-then-insert per source […] splitting across overlapping date ranges lets a later chunk's pre-cleanup wipe an earlier chunk's inserts.

> A key that lands on a different day than the server's lets one chunk's pre-cleanup range overlap another chunk's days and delete its freshly inserted rows.

That mitigation lives entirely in the client. The server has no defence of its own.

**7. Concurrency is confirmed empirically.** Two failures landed 1 ms apart (`05:16:51.444` / `.445`). In-request processing is a sequential `for` loop and each `createExerciseEntry` performs several round-trips, so two of them cannot complete 1 ms apart within one request. Overlapping requests are therefore certain.

**Two hypotheses I explored and had to retract** — recording them so nobody re-treads them:

- *Delegated-access mismatch.* `can_access_user_data` has an inheritance clause (`rls_policies.sql:371-374`) letting a `can_view_reports`-only delegate pass `checkPermissionMiddleware('diary')` on a write, which RLS then denies. This is a genuine latent route-guard/RLS inconsistency worth fixing separately, but it is **not** this bug — the account has no delegation at all.
- *A `v1.6.2` regression in `resolveHealthEntryDate`.* `v1.6.2` did drop `&& !entry.timestamp` from `basisIsDayOnly`. But `ExerciseSession`/`Workout` take a dedicated earlier branch (`basisField = entry.timestamp || entry.date || entry.entry_date`) that is present in **all** of `v1.6.1`–`v1.6.4`, so `basisField` is always a full instant and `basisIsDayOnly` is always false for workouts. That diff never affected this path. **I could not bisect a specific regressing commit** — the structure described above predates the 1.6.x line.

---

## Issues

**Issue 1 — A missing parent row is reported as a permissions error.** *(confirmed)*

`exercise_entry_sets` and `exercise_entry_activity_details` are the only exercise tables whose RLS derives from a parent lookup rather than their own `user_id` (`rls_policies.sql:673-677`, `:865-877`):

```sql
WITH CHECK (EXISTS (SELECT 1 FROM public.exercise_entries ee
  WHERE ee.id = exercise_entry_sets.exercise_entry_id AND has_diary_access(ee.user_id)))
```

Any absent parent — for any reason — surfaces as `42501` instead of a foreign-key violation, because `WITH CHECK` is evaluated before the FK trigger fires. This is a diagnosability defect independent of the race below.

**Issue 2 — Sync pre-cleanup, parent write, and child writes are not atomic.** *(confirmed)*

Three separate transactions on three separate connections, with `READ COMMITTED` snapshots per statement. A concurrent commit that range-deletes the same `(source, day-span)` can remove the parent after it has been observed and before its children are written. `_updateExerciseEntryWithClient` does guard existence up front (`SELECT * FROM exercise_entries WHERE id = $1 AND user_id = $2`, throwing `Exercise entry not found for update.`), but several statements run between that guard and the child insert, so the guard is not a guarantee.

**Issue 3 — The parent `UPDATE` result is not checked.** *(confirmed, minor)*

`UPDATE exercise_entries SET … WHERE id = $28 AND user_id = $29 RETURNING id` — the returned row is not inspected. If the row vanishes after the guard, the update quietly affects zero rows and execution proceeds to write children against an id that no longer exists.

**Issue 4 — The concurrent deleter is not identified.** *(hypothesis — needs maintainer input)*

Overlapping requests are proven, but not their origin. Candidates:

- Two workout chunks in flight simultaneously, or a chunk retry (`healthDataApi.ts` retries whole chunks) re-running a destructive pre-cleanup while the first attempt is still processing server-side — plausible if a reverse proxy times out a telemetry-heavy chunk while the server keeps working.
- A scheduled server-side provider sync overlapping the mobile sync. Notably the failing session originated in Hevy (`@hevyapp` in the notes) and arrived tagged `source: 'Health Connect'`, while `scheduleHevySyncs()` also runs on cron — two writers over the same activity.
- Any second client (web) triggering a re-sync of the same range.

**Issue 5 — Route-guard/RLS permission mismatch.** *(confirmed, latent, unrelated to this report)*

`can_access_user_data(target, 'diary', auth)` returns true for a delegate holding only `can_view_reports`/`reports`, so such a caller passes the write guard on `POST /health-data` and is then denied by `has_diary_access`, which requires `can_manage_diary`. Filed here for visibility; it would produce the same `42501` with a genuinely different cause.

---

## Proposed solution

Deliberately separate from the diagnosis above; maintainers may well prefer a different split.

**Fix 1 — Make one sync request atomic (addresses Issue 2, the actual failure).**

Thread a single client and transaction through pre-cleanup → `exercise_entries` → `exercise_entry_sets` → `exercise_entry_activity_details` for a request. `*WithClient` variants of all three already exist (`deleteExerciseEntriesByEntrySourceAndDateWithClient`, `_createExerciseEntryWithClient`, `_createActivityDetailWithClient`), so this is largely a wiring change in `measurementService.processHealthData` and `healthDataHandlers.workoutHandler` rather than new logic. Once parent and children share a transaction, a competing delete cannot land between them.

**Fix 2 — Serialise destructive range-deletes per user+source (addresses Issue 4 without needing to identify the writer).**

Take a `pg_advisory_xact_lock` keyed on `(user_id, source)` for the span of a sync request, so two overlapping syncs of the same source queue instead of interleaving delete-then-insert. This is robust whether the second writer is a retried chunk, a parallel chunk, or a provider cron.

**Fix 3 — Report a missing parent honestly (addresses Issues 1 and 3).**

Check the `RETURNING id` result of the parent `UPDATE` and fail with a clear "parent entry no longer exists" error before attempting child writes. Optionally re-assert parent existence immediately before the child insert. This converts a confusing `42501` into an accurate, actionable error and makes any residual race obvious in logs instead of looking like an RLS misconfiguration.

**Fix 4 — Make the pre-cleanup non-destructive to rows it is about to re-insert (optional hardening).**

Restrict the range-delete to entries whose `source_id` is *not* present in the incoming batch, so a retried or overlapping chunk cannot destroy rows that a concurrent request just wrote.

**Fix 5 — Align the route guard with RLS (addresses Issue 5, separable).**

Drop `'diary'` from the read-inheritance list in `can_access_user_data`, or resolve write requests to a `diary_write` permission that maps to `can_manage_diary`, so the route guard and `has_diary_access` agree.

### Steps to reproduce

1. Configure the Android mobile app to sync Health Connect workout data, with enough workout history that the payload exceeds the 2 MB workout chunk soft limit (or that a chunk request is slow enough for the client to retry).
2. Trigger a Health Connect sync covering a date range that another writer touches concurrently — e.g. a scheduled provider sync for the same activities, or a second sync started before the first finishes.
3. Watch the server log.

Reliable direct reproduction of the race, if you prefer determinism: begin a `createExerciseEntry` for an existing provider-sourced entry, and from a second connection commit `deleteExerciseEntriesByEntrySourceAndDate` for that user/source/day between the entry's existence check and its sets insert. The sets insert fails with `42501`.

### Expected behavior

Health Connect workout sync stores sets and activity details without error. If a parent entry genuinely disappears mid-request, the server should report that plainly (or avoid it entirely by keeping parent and children in one transaction) rather than emitting a row-level-security violation that implies a permissions misconfiguration.

Docker Compose

### SparkyFitness version

v1.6.4

### Which area is affected?

Server / API, Database / migrations, Mobile app (Android), Integrations (Garmin, Google Health, Polar, Oura, etc.)

```shell
[2026-09-01T05:16:51.444Z] [ERROR] Error creating/updating exercise entry with snapshot: error: new row violates row-level security policy for table "exercise_entry_sets"
    at async _updateExerciseEntryWithClient (/app/SparkyFitnessServer/models/exerciseEntry.ts:581:7)
    at async _createExerciseEntryWithClient (/app/SparkyFitnessServer/models/exerciseEntry.ts:655:28)
    at async Object.createExerciseEntry (/app/SparkyFitnessServer/models/exerciseEntry.ts:786:23)
    at async Object.handle (/app/SparkyFitnessServer/services/healthDataHandlers.ts:1522:29)
    at async Object.processHealthData (/app/SparkyFitnessServer/services/measurementService.ts:326:23)
    at async <anonymous> (/app/SparkyFitnessServer/integrations/healthData/healthDataRoutes.ts:60:20) {
  severity: 'ERROR',
  code: '42501',
  file: 'execMain.c',
  line: '2337',
  routine: 'ExecWithCheckOptions'
}
[2026-09-01T05:16:51.445Z] [ERROR] Error creating/updating exercise entry with snapshot: error: new row violates row-level security policy for table "exercise_entry_sets"
    (identical frames — note the 1 ms gap, evidencing overlapping requests)

[2026-09-01T05:18:08.725Z] [ERROR] Failed to create activity detail for entry ID 3b76632d-5072-40df-996a-f7a2fbc414aa: new row violates row-level security policy for table "exercise_entry_activity_details" {
  query: '\n' +
    '        INSERT INTO exercise_entry_activity_details (\n' +
    '            exercise_entry_id,\n' +
    '            exercise_preset_entry_id,\n' +
    '            provider_name,\n' +
    '            detail_type,\n' +
    '            detail_data,\n' +
    '            created_by_user_id,\n' +
    '            updated_by_user_id\n' +
    '        )\n' +
    '        VALUES ($1, $2, $3, $4, $5, $6, $7)\n' +
    '        RETURNING *;\n' +
    '    ',
  values: [
    '3b76632d-5072-40df-996a-f7a2fbc414aa',
    undefined,
    'Health Connect',
    'ExerciseSession_raw_data',
    {
      exerciseType: 70,
      title: 'Calisthenics Push (Cut)',
      segments: [],
      notes: 'Calisthenics Push (Cut)\nSaturday, Aug 29, 2026 at 6:08pm\n ... \n@hevyapp\nhttps://hevy.com/workout/...',
      endTime: '2026-08-30T02:51:14Z',
      startTime: '2026-08-30T01:08:48Z'
    },
    '<user-id>',
    '<user-id>'
  ],
  error: error: new row violates row-level security policy for table "exercise_entry_activity_details"
      at async _createActivityDetailWithClient (/app/SparkyFitnessServer/models/activityDetailsRepository.ts:47:20)
      at async Object.createActivityDetail (/app/SparkyFitnessServer/models/activityDetailsRepository.ts:70:12)
      at async Object.handle (/app/SparkyFitnessServer/services/healthDataHandlers.ts:1563:9)
      at async Object.processHealthData (/app/SparkyFitnessServer/services/measurementService.ts:326:23)
      at async <anonymous> (/app/SparkyFitnessServer/integrations/healthData/healthDataRoutes.ts:60:20) {
    severity: 'ERROR',
    code: '42501',
    routine: 'ExecWithCheckOptions'
  }
}

[2026-09-01T05:19:10.500Z] [ERROR] Failed to create activity detail for entry ID 8e01fc6b-117b-4f3b-acf1-656695c637db: new row violates row-level security policy for table "exercise_entry_activity_details"
    (same shape; detail_data title 'WALKING', exerciseType 79,
     startTime '2026-08-14T15:48:22Z', endTime '2026-08-14T16:05:54Z')
```

### Anything else?

Diagnostic queries run against the live database, and their results, so the permissions angle need not be re-checked:

```sql
-- Parent row referenced by the failing activity-detail insert: ABSENT
SELECT id, user_id, created_by_user_id, source FROM exercise_entries
WHERE id = '8e01fc6b-117b-4f3b-acf1-656695c637db';
-- (0 rows)

-- Delegation for the syncing user: NONE (so has_diary_access short-circuits true on own rows)
SELECT owner_user_id, family_user_id, is_active, access_end_date, access_permissions
FROM family_access WHERE family_user_id = '<user-id>';
-- (0 rows)

-- Policies: all present and correct
SELECT tablename, policyname, cmd FROM pg_policies
WHERE tablename IN ('exercise_entry_sets','exercise_entry_activity_details','exercise_entries');
-- exercise_entry_activity_details | select_policy                              | SELECT
-- exercise_entry_activity_details | modify_policy                              | ALL
-- exercise_entry_sets             | select_policy                              | SELECT
-- exercise_entry_sets             | modify_policy                              | ALL
-- exercise_entries                | select_policy                              | SELECT
-- exercise_entries                | modify_policy                              | ALL
-- exercise_entries                | select_exercise_preset_entry_linked_policy | SELECT
```

Also checked and ruled out:

- `rls_policies.sql` is applied as a single multi-statement `client.query()` with no transaction control and no existence guards, so it is all-or-nothing — partial policy application is not possible, and the startup log shows no `Error applying RLS policies`.
- `poolManager.ts` and `authMiddleware.ts` are byte-identical from `v1.6.1` to `main`; `getClient()` correctly sets `app.authenticated_user_id` via `set_app_context` on every checkout.
- The `exercise_entry_sets` / `exercise_entry_activity_details` policy text is character-for-character identical from `v1.6.1` through `main`; only line numbers shifted. Upgrading past `v1.6.4` will not change this behaviour.

jsandai:
## Description

Health Connect workout sync can lose a workout's sets and its raw provider data while reporting success. Both child writes fail with a row level security error because the parent `exercise_entries` row was deleted by a concurrent sync of the same source, and neither the parent `UPDATE` result nor the separate activity detail write was guarded against that.

Two changes, both in the sync write path.

1. The parent `UPDATE` result is now checked. A zero row update used to pass silently and the child inserts then ran against an id that no longer existed. On the sync path a vanished parent means "insert", since delete-then-insert ingest is idempotent, so it re-runs its own deduplication lookup once and updates the replacement row if the competing writer already committed one, and inserts only when the key is genuinely free. Callers acting on a user chosen entry still get a real error.
2. The raw activity detail is written on the entry's own connection inside the same transaction, through the existing `_createActivityDetailWithClient`, instead of afterwards on a connection of its own. The parent is then present and locked when the child insert runs.

The reason this surfaced as a permissions error rather than a foreign key violation is that both child tables resolve their parent through an `EXISTS` subquery in their RLS policy, and `WITH CHECK` is evaluated before the foreign key's `AFTER ROW` trigger.

Linked Issue: Closes #2351

## How to Test

1. Check out this branch, then `cd SparkyFitnessServer && pnpm install`.
2. Run `pnpm run typecheck && pnpm run lint && pnpm run format:check`.
3. Run `pnpm run test:ci`. The new coverage is in `tests/exerciseEntry.vanishedParent.test.ts` and `tests/exerciseEntry.activityDetail.test.ts`, plus two cases added to `tests/workoutHandlerTelemetry.test.ts`.
4. For the behaviour itself: sync a Health Connect workout batch that carries sets and raw activity data, then confirm the rows land in `exercise_entry_sets` and `exercise_entry_activity_details` and that no `42501` appears in the Postgres log.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

No UI changes, so there is nothing to show.

## Notes for Reviewers

There are no new tables and no new endpoints, so the Zod and `rls_policies.sql` items above are satisfied by there being nothing to add. No migrations either.

A known limit, and the part I would most like a second opinion on: the re-check narrows the duplicate window but it cannot close it: if our lookup lands after the competing sync's delete commits and before its re-insert commits, we insert and it inserts, and there is no unique constraint on `(user_id, source, source_id)` to reject either row. Closing it properly needs either that unique key plus an upsert, which wants a dedupe migration first because existing installs may already hold duplicates from this bug, or a lock held across the whole delete-then-insert group rather than inside a single entry write. Both felt like a call @CodeWithCJ should make rather than something to slip into a bug fix. The trade as it stands is that the old code failed loudly and lost the workout, and this keeps the workout at the cost of a possible duplicate that clears on the next sync of that day.

Two smaller things left alone deliberately. The pre-cleanup still commits in its own transaction before per record validation, so a failure part way through a batch leaves the range delete committed. And the other provider integrations (Oura, Hevy, Strava, Google Health, Fitbit, Polar, Withings) still store their raw dumps on a separate connection after the entry write, so they carry the same latent race. Garmin and the FIT importer already write in transaction. Adopting the new option is about one line each if you want that in a follow up.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workout synchronization when entries are deleted and recreated concurrently, preventing updates to stale records.
  - Prevented duplicate workouts during overlapping synchronization operations.
  - Preserved the original activity source when updating exercise entries.
  - Saved provider activity details atomically with the associated workout, rolling back both if saving fails.
  - Replaced outdated activity details when updated workout data is synchronized.
  - Improved handling of missing entries during synchronization to maintain consistent workout data.

### DIFF
--- a/SparkyFitnessServer/models/activityDetailsRepository.ts
+++ b/SparkyFitnessServer/models/activityDetailsRepository.ts
@@ -1,5 +1,6 @@
 import { getClient } from '../db/poolManager.js';
 import { log } from '../config/logging.js';
+import type { PoolClient } from 'pg';
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 async function _createActivityDetailWithClient(client: any, detail: any) {
   const {
@@ -297,24 +298,27 @@ async function deleteActivityDetail(userId: any, id: any) {
   }
 }
 async function _deleteActivityDetailsByEntryIdAndProviderWithClient(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  client: any,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  userId: any,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  entryId: any,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  providerName: any
+  client: PoolClient,
+  userId: string,
+  entryId: string,
+  providerName: string,
+  detailType?: string
 ) {
   const query = `
         DELETE FROM exercise_entry_activity_details
         WHERE (exercise_entry_id = $1 OR exercise_preset_entry_id = $1)
           AND provider_name = $2
+          AND ($4::text IS NULL OR detail_type = $4)
           AND (exercise_entry_id IN (SELECT id FROM exercise_entries WHERE user_id = $3)
                OR exercise_preset_entry_id IN (SELECT id FROM exercise_preset_entries WHERE user_id = $3));
     `;
   try {
-    const result = await client.query(query, [entryId, providerName, userId]);
+    const result = await client.query(query, [
+      entryId,
+      providerName,
+      userId,
+      detailType ?? null,
+    ]);
     log(
       'debug',
       `Successfully deleted ${result.rowCount} activity details for entry ID ${entryId} and provider ${providerName}.`
--- a/SparkyFitnessServer/models/exerciseEntry.ts
+++ b/SparkyFitnessServer/models/exerciseEntry.ts
@@ -605,6 +605,12 @@ async function _updateExerciseEntryWithClient(
   }
   return _getExerciseEntryByIdWithClient(client, id);
 }
+// _updateExerciseEntryWithClient hands back the whole entry row, but the create
+// path below only needs the identity of the row it touched: the full entry is
+// re-read through _getExerciseEntryByIdWithClient before being returned.
+interface MatchedExerciseEntry {
+  id: string;
+}
 async function _createExerciseEntryWithClient(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   client: any,
@@ -631,39 +637,45 @@ async function _createExerciseEntryWithClient(
       'Fitbit',
       'Strava',
     ].includes(entrySource);
-    let existingEntryResult;
-    // 1. Attempt precise sync deduplication via source_id if available
-    if (syncDuplicateCheck) {
-      existingEntryResult = await client.query(
-        'SELECT id FROM exercise_entries WHERE user_id = $1 AND source = $2 AND source_id = $3',
-        [userId, entrySource, entryData.source_id]
-      );
-    }
-    // 2. If no source_id match and NOT a sync source, fall back to "Manual" deduplication (name/date).
-    // Skip this fallback when source_id was provided: a source_id miss means it's a genuinely new
-    // activity (different activityId), so we must INSERT rather than match on exercise_id + date.
-    if (
-      !existingEntryResult?.rows?.length &&
-      !exercisePresetEntryId &&
-      !skipManualDuplicateCheck &&
-      !syncDuplicateCheck &&
-      !options.skipDuplicateCheck
-    ) {
-      if (entryData.workout_plan_assignment_id) {
-        // If it's linked to a workout plan assignment, it's unique by that assignment ID and date.
+    // Both deduplication lookups live behind one function so that the update
+    // path can re-run exactly the lookup that produced its match. Returns the
+    // id of the matched entry, or null when this is a genuinely new one.
+    const findExistingEntryId = async (): Promise<string | null> => {
+      let existingEntryResult;
+      // 1. Attempt precise sync deduplication via source_id if available
+      if (syncDuplicateCheck) {
         existingEntryResult = await client.query(
-          'SELECT id FROM exercise_entries WHERE user_id = $1 AND workout_plan_assignment_id = $2 AND entry_date = $3',
-          [userId, entryData.workout_plan_assignment_id, entryData.entry_date]
-        );
-      } else {
-        // For manual entries (no assignment), keep traditional uniqueness check by exercise_id and date.
-        // We explicitly ensure workout_plan_assignment_id is NULL to avoid matching template-generated entries.
-        existingEntryResult = await client.query(
-          'SELECT id FROM exercise_entries WHERE user_id = $1 AND exercise_id = $2 AND entry_date = $3 AND source = $4 AND exercise_preset_entry_id IS NULL AND workout_plan_assignment_id IS NULL',
-          [userId, entryData.exercise_id, entryData.entry_date, entrySource]
+          'SELECT id FROM exercise_entries WHERE user_id = $1 AND source = $2 AND source_id = $3',
+          [userId, entrySource, entryData.source_id]
         );
       }
-    }
+      // 2. If no source_id match and NOT a sync source, fall back to "Manual" deduplication (name/date).
+      // Skip this fallback when source_id was provided: a source_id miss means it's a genuinely new
+      // activity (different activityId), so we must INSERT rather than match on exercise_id + date.
+      if (
+        !existingEntryResult?.rows?.length &&
+        !exercisePresetEntryId &&
+        !skipManualDuplicateCheck &&
+        !syncDuplicateCheck &&
+        !options.skipDuplicateCheck
+      ) {
+        if (entryData.workout_plan_assignment_id) {
+          // If it's linked to a workout plan assignment, it's unique by that assignment ID and date.
+          existingEntryResult = await client.query(
+            'SELECT id FROM exercise_entries WHERE user_id = $1 AND workout_plan_assignment_id = $2 AND entry_date = $3',
+            [userId, entryData.workout_plan_assignment_id, entryData.entry_date]
+          );
+        } else {
+          // For manual entries (no assignment), keep traditional uniqueness check by exercise_id and date.
+          // We explicitly ensure workout_plan_assignment_id is NULL to avoid matching template-generated entries.
+          existingEntryResult = await client.query(
+            'SELECT id FROM exercise_entries WHERE user_id = $1 AND exercise_id = $2 AND entry_date = $3 AND source = $4 AND exercise_preset_entry_id IS NULL AND workout_plan_assignment_id IS NULL',
+            [userId, entryData.exercise_id, entryData.entry_date, entrySource]
+          );
+        }
+      }
+      return existingEntryResult?.rows?.[0]?.id ?? null;
+    };
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     let newEntryId: any;
     let operation: 'created' | 'updated';
@@ -672,10 +684,9 @@ async function _createExerciseEntryWithClient(
     // overlapping sync of the same source is enough. For delete-then-insert
     // ingest a vanished parent means "insert", not "fail", so fall through to
     // the create path below rather than writing children against a dead id.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    let updatedEntry: any = null;
-    if (existingEntryResult && existingEntryResult.rows.length > 0) {
-      const existingEntryId = existingEntryResult.rows[0].id;
+    let updatedEntry: MatchedExerciseEntry | null = null;
+    const existingEntryId = await findExistingEntryId();
+    if (existingEntryId) {
       log(
         'info',
         `Existing exercise entry found for user ${userId}, exercise ${entryData.exercise_id}, date ${entryData.entry_date}, source ${entrySource}. Updating entry ${existingEntryId}.`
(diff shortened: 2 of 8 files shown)


jsandai:
@coderabbitai review

jsandai:
@coderabbitai full review

CodeWithCJ (approved):

PR #2368, opened 2026-09-03
fix(exercise): cache the free-exercise-db dataset

### LINKED ISSUE #2367: [Bug]: free-exercise-db search re-downloads the full dataset on every distinct query and exhausts the GitHub API rate limit

Yes, I'd like to submit a PR!

### What happened?

Exercise search against the free-exercise-db provider downloads the entire exercises.json (about 1 MB) from the GitHub REST API on every distinct search term, rather than caching the dataset and filtering it in memory.

The response is cached, but the cache key is built from the query, the filters, the limit and the offset, so a new search term is always a miss. Each miss issues a fresh request to:

  https://api.github.com/repos/yuhonas/free-exercise-db/contents/dist/exercises.json

That endpoint is the unauthenticated GitHub REST API, which allows 60 requests per hour per IP. The limit is per IP rather than per user, so it's shared by everyone on an instance. Once it's exhausted GitHub returns 403 and exercise search silently returns no results for the rest of the hour.

free-exercise-db is seeded as a default provider for every user, so this is on the normal path rather than something only an unusual configuration hits. A single active user can exhaust the hourly budget on their own, and a household or small shared instance will do it quickly.

The file already defines a GITHUB_RAW_BASE_URL constant pointing at raw.githubusercontent.com and uses it in getExerciseById. Only searchExercises reaches for the API host, which is the one subject to the 60 per hour limit.

### Steps to reproduce

1. Use an account with free-exercise-db enabled as an exercise provider (it's a default, so no setup is needed).
2. Open exercise search and switch to the Online tab with free-exercise-db selected.
3. Search for a term, for example "squat".
4. Search for a different term, for example "bench".
5. Watch the server logs. Each distinct term logs "[FreeExerciseDBService] Fetching exercises from: https://api.github.com/..." and pulls the whole dataset again.
6. Repeat with more than 60 distinct terms within an hour, or leave a couple of users searching normally. Further searches return empty results, and the server logs a 403 from GitHub.

### Expected behavior

The dataset should be downloaded once, cached, and filtered in memory, so the number of upstream requests is a function of the cache lifetime rather than of how many different things people search for. Searching should keep working regardless of how many distinct terms are used.

Docker Compose

### SparkyFitness version

1.6.4

### Which area is affected?

Server / API, Integrations (Garmin, Google Health, Polar, Oura, etc.)

```yaml

```

```shell
[FreeExerciseDBService] Fetching exercises from: https://api.github.com/repos/yuhonas/free-exercise-db/contents/dist/exercises.json

(repeats once per distinct search term; after the hourly budget is used up the request fails with a 403 and searchExercises returns an empty result set)
```

_No response_

### Anything else?

Two things make this cheap to fix.

The dataset is small. exercises.json is about 1 MB, so a single parsed copy per process is enough and there's no need for external cache infrastructure.

The licence is not a constraint. yuhonas/free-exercise-db is released under The Unlicense, a public domain dedication, and its README explicitly says the JSON can be used locally. The limit being hit here is the GitHub API rate limit rather than anything to do with the data, and caching reduces the number of calls, so it moves us further inside GitHub's terms rather than closer to the edge.

jsandai:
## Description

The free-exercise-db integration re-downloads the entire `exercises.json` on every distinct search term, because the cache key includes the query, the filters, the limit and the offset. It also fetches through the GitHub REST API, which is the unauthenticated 60 requests per hour bucket, counted per IP and so shared by everyone on an instance. free-exercise-db is seeded as a default provider for every user, so this is on the common path.

The parsed dataset is now held once in a module-level record with its fetch timestamp and filtered in memory, fetched from `raw.githubusercontent.com` using the `GITHUB_RAW_BASE_URL` constant that was already in the file. It's deliberately kept out of the shared `NodeCache`, whose `useClones` default would otherwise deep-clone the whole dataset on every read. A single-flight promise means a burst of cold searches after a restart triggers one download rather than one per search.

The refresh path is bounded and defensive: the request carries an explicit timeout so a stalled connection can't leave every search waiting on one shared promise, the response is validated as an exercise array before it's cached so a proxy interstitial or error page can't poison the cache for the full TTL, and a failed refresh serves the last good dataset with a warning that includes its age. A five minute retry interval then applies whether or not a cached dataset exists, so a cold-start outage fails fast instead of firing a fresh request per search.

`searchExercises()` keeps its signature, so no caller changes.

Linked Issue: Closes #2367

## How to Test

1. Check out this branch and run `pnpm test` in `SparkyFitnessServer`.
2. The tests in `tests/freeExerciseDBSearch.test.ts` assert the new behaviour directly: two different queries result in exactly one `axios.get`, concurrent cold searches share one download, the request goes to the raw host, a timeout is configured, an invalid payload is rejected rather than cached, and a failed refresh serves the previous dataset and then backs off for the retry interval.
3. For a manual check, search the Online tab with free-exercise-db selected, then search again with a different term, and confirm only the first search hits the network.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

Not applicable

## Notes for Reviewers

There were no tests asserting how often the network gets hit, which is why the per-query key went unnoticed for so long, so the new tests cover that specifically. The file goes from 2 tests to 14.

The dataset is held outside `NodeCache` on purpose. `node-cache` defaults `useClones` to true, so keeping a multi-megabyte dataset in it would deep-clone the whole thing on every read, which measured at roughly 6ms per search locally. The `getExerciseById` cache entry is unchanged and still uses `NodeCache`.

Stale data is served with a warning rather than silently, so an instance whose upstream is permanently failing shows up in the logs instead of quietly serving very old exercises.

## Summary by CodeRabbit

- **Performance**
  - Exercise searches now reuse a cached dataset, reducing repeated downloads and improving response times.
  - Concurrent searches share a single data request instead of triggering duplicate downloads.

- **Reliability**
  - Added request timeouts and automatic retry handling for unavailable or invalid exercise data.
  - Previously retrieved data remains available during temporary refresh failures.

- **Data Quality**
  - Exercise datasets are fully validated before use, preventing malformed or inconsistent entries from appearing in search results.

### DIFF
--- a/SparkyFitnessServer/integrations/freeexercisedb/FreeExerciseDBService.ts
+++ b/SparkyFitnessServer/integrations/freeexercisedb/FreeExerciseDBService.ts
@@ -27,7 +27,7 @@ interface DatasetHolder {
 }
 
 function isExerciseDataset(value: unknown): value is FreeExercise[] {
-  if (!Array.isArray(value)) return false;
+  if (!Array.isArray(value) || value.length === 0) return false;
   return value.every(
     (exercise: unknown) =>
       typeof exercise === 'object' &&
@@ -94,10 +94,6 @@ class FreeExerciseDBService {
       now - lastDatasetFetchFailureAt < STALE_RETRY_INTERVAL_MS
     ) {
       if (dataset) {
-        log(
-          'warn',
-          `[FreeExerciseDBService] Serving stale exercise dataset after a refresh failure; age: ${now - dataset.fetchedAt}ms`
-        );
         return dataset.data;
       }
       log(
@@ -132,13 +128,11 @@ class FreeExerciseDBService {
           }
           if (dataset) {
             const age = Date.now() - dataset.fetchedAt;
-            if (age >= DATASET_TTL_MS) {
-              log(
-                'warn',
-                `[FreeExerciseDBService] Serving stale exercise dataset after a refresh failure; age: ${age}ms`,
-                error instanceof Error ? error.message : error
-              );
-            }
+            log(
+              'warn',
+              `[FreeExerciseDBService] Serving stale exercise dataset after a refresh failure; age: ${age}ms`,
+              error instanceof Error ? error.message : error
+            );
             return dataset.data;
           }
           throw error;
--- a/SparkyFitnessServer/integrations/freeexercisedb/FreeExerciseDBService.ts
+++ b/SparkyFitnessServer/integrations/freeexercisedb/FreeExerciseDBService.ts
@@ -8,6 +8,41 @@ const GITHUB_RAW_BASE_URL =
 const EXERCISES_PATH = 'exercises'; // No leading slash for API
 // Initialize cache for GitHub API responses (e.g., 1 hour TTL)
 const githubCache = new NodeCache({ stdTTL: 3600 });
+const DATASET_TTL_MS = 60 * 60 * 1000;
+// Bound a stalled shared download so it cannot block every exercise search indefinitely.
+const DATASET_REQUEST_TIMEOUT_MS = 15 * 1000;
+// Avoid hammering GitHub and making every search wait during an upstream outage.
+const STALE_RETRY_INTERVAL_MS = 5 * 60 * 1000;
+
+interface FreeExercise {
+  name: string;
+  equipment?: string;
+  primaryMuscles?: string[];
+  secondaryMuscles?: string[];
+}
+
+interface DatasetHolder {
+  data: FreeExercise[];
+  fetchedAt: number;
+}
+
+function isExerciseDataset(value: unknown): value is FreeExercise[] {
+  if (!Array.isArray(value)) return false;
+  if (value.length === 0) return true;
+
+  const firstExercise: unknown = value[0];
+  return (
+    typeof firstExercise === 'object' &&
+    firstExercise !== null &&
+    'name' in firstExercise &&
+    typeof firstExercise.name === 'string'
+  );
+}
+
+let dataset: DatasetHolder | null = null;
+let exercisesDatasetPromise: Promise<FreeExercise[]> | null = null;
+let lastDatasetFetchFailureAt: number | null = null;
+
 class FreeExerciseDBService {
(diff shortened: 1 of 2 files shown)


jsandai:
@coderabbitai

Fixed in the latest commit. `isExerciseDataset` now requires a string `name` on every entry rather than only the first, so a mixed payload like `[{ name: 'Lunge' }, {}]` is rejected instead of being cached for the hour. There's a regression test covering exactly that shape, plus one confirming a rejected payload doesn't poison the cache: after the retry interval a later search hits the network again and returns real results.

I've deliberately left `equipment`, `primaryMuscles` and `secondaryMuscles` out of the validation. They're optional on the upstream records, and the filter path already tolerates their absence: `equipment` is read with optional chaining, and both muscle arrays are truthiness-guarded before use. So unlike `name`, they can't throw. Requiring them would reject legitimate entries, since plenty of exercises in the dataset have no equipment listed. I've added a test that pins that decision, so a later change can't quietly tighten the check and start dropping valid data.

CodeWithCJ:
One thing before merge:

`SparkyFitnessServer/integrations/freeexercisedb/FreeExerciseDBService.ts:29` —
isExerciseDataset accepts an empty array, since [].every() is vacuously true. If
the upstream ever serves `[]`, it caches as the good dataset for the full hour
with no warning and search silently returns nothing. Unlikely, but it is the one
bad-response shape the validator lets through, and `value.length > 0 &&` closes it.

Non-blocking: the `age >= DATASET_TTL_MS` check at line 135 can never be false —
line 89 only falls through once the dataset is already past the TTL, and nothing
refreshes it between there and the catch. Harmless, but it reads as a guard that
does something. And every search during the 5-minute retry window logs a warn at
line 97; once per window would be enough on a busy instance.

The rest reads correctly — I traced the single-flight identity guard and the
cold-start, stale-serve and timeout paths.

CodeWithCJ (approved):

Issue #2269, opened 2026-08-26
[Bug]: workout rest timer sound does not play if the app is in the background


Jerrys-modz:
### Are you willing to submit a PR?

No, but I can help test

### What happened?

Hi the workout rest timer sound does not play if you have the app in the background. It only plays when you have the app currently open with screen on. Would be nice if I could have my phone in my pocket and hear the sound 

I am on 1.6.3 on iOS

### Steps to reproduce

Start a work out and finish a set with a rest timer. Open a different app or lock your phone and no sound will play just the notification will show up

### Expected behavior

The rest timer sound will play 

### How are you running SparkyFitness?

Docker Compose

### SparkyFitness version

1.6.3

### Which area is affected?

Mobile app (iOS)

### Docker Compose & Environment Configuration

```yaml

```

### Runtime Logs (Server, Web, or Mobile)

```shell

```

almottier:
There's the same issue on Android.

zeusal:
> There's the same issue on Android.

On iOS I found the cause: the rest alert was scheduled at the default `active` interruption level, which any Focus mode (and the Scheduled Summary) delivers silently — the notification lands on the Lock Screen but never makes a sound. On top of that the app's own notification handler muted the ping in states where the in-app chime doesn't play either, so both cues cancelled each other out. That fix is iOS-only by construction, so it won't do anything for Android.

Reading the Android path I can't explain the silence from the code alone: a backgrounded rest ping is presented with no behaviour override, so `sound: true` falls through to the "Workout timer" channel's own sound, and the alarm is scheduled as exact-and-allow-while-idle, so Doze shouldn't swallow it outright. To pin down what's actually happening on your device:

1. Does the notification appear at the moment the rest ends, or noticeably late?
2. Is Do Not Disturb, Bedtime mode, or a similar manufacturer mode active when it happens?
3. Settings → Apps → SparkyFitness → Notifications → "Workout timer": what is Sound set to, and is the importance still the highest option?
4. Settings → Apps → Special app access → Alarms & reminders: is SparkyFitness allowed? (If you ever tapped "Not Now" on that prompt, it isn't shown again.)
5. Does the sound play if the app is open on screen when the rest ends?
6. Android version and phone model?

If it turns out to be Do Not Disturb, the Android fix is a different one from the iOS one — the channel would need `setBypassDnd`, which requires the Do Not Disturb access permission — so it's probably worth splitting into its own issue either way.

[...]

PR #2384, opened 2026-09-06
fix(mobile): rest timer background sound

zeusal:
## Description

The workout rest timer only makes a sound while the app is open on screen: lock the phone or switch apps and the rest-complete notification appears without a sound. Three separate things in the rest-cue path can silence it.

The rest timer has two cues that are meant to be mutually exclusive — the in-app chime while the app is on screen, the scheduled ping's notification sound otherwise — but both sides decided from the `restTimerSoundEnabled` preference alone, and the ping went out at the default interruption level.

- **Notification handler.** iOS runs it for anything delivered while the app is frontmost, including when the app is merely *inactive* (screen locking, app switcher, Control Center, an incoming call). `playRestCompleteSound` refuses to play there, yet the handler still muted the ping as a duplicate and hid its banner, so the rest landed with no sound and nothing to look at. The ping is now muted only when the chime will really play, and is otherwise presented like the background delivery it effectively is. `willPlayRestCompleteSound()` holds that "preference **and** `AppState`" test so the two sides cannot drift apart again.
- **Deadline flip.** `markRestReady` cancelled the pending ping at the exact instant the OS was due to deliver it. The deadline timer runs in those same inactive states, so the cancel either loses the race or removes the only cue the user was going to get. It now cancels only while the app is active, where the chime and haptic already cover the flip.
- **Interruption level.** The ping was scheduled at the default `active` level, which any Focus mode (and the Scheduled Summary) delivers *silently* — the alert still lands on the Lock Screen, it just never makes a sound. That is exactly the reported symptom, and Focus is on by default during Sleep and easy to leave running. The rest alert is now `timeSensitive`, with the `com.apple.developer.usernotifications.time-sensitive` entitlement that lets iOS honour it. iOS falls back to `active` when the entitlement is missing and Android ignores the field, so there is no regression risk in the fallback path.

**Platform scope — this fixes iOS only.** All three changes are iOS-specific and inert (not harmful) on Android: `interruptionLevel` is an iOS-only field, with no reference to it anywhere under `expo-notifications/android/`, and `AppState`'s `inactive` state — the window the handler and deadline-flip fixes target — is documented by React Native as iOS-only.

The issue also carries a report of the same symptom on Android, and **this PR does not claim to fix that**, because it turned out to be a different problem entirely. An Android user on the thread traced it to keeping the phone on silent: the app configures expo-audio with `playsInSilentMode: false`, and that is exactly what expo-audio's `RingerModeReceiver` consults before pausing every player the moment the ringer leaves `RINGER_MODE_NORMAL`. Nothing in the code paths fixed here is involved. That half is addressed separately in #2506, for the in-app chime while the app is on screen; an alert arriving with the app closed remains subject to the device's silent mode on both platforms, which would need Critical Alerts on iOS or a Do Not Disturb bypass on Android to change.

Linked Issue: #2269 — deliberately not `Closes`, since the thread also carries the Android report described above, which this PR does not address.

## How to Test

1. Check out this branch, then from `SparkyFitnessMobile/`: `npx expo prebuild --clean` and build to a device (the entitlement is a native change; a JS-only reload will not pick it up).
2. Settings → Workout Settings: leave **Rest timer sound** on. Settings → Notifications: leave **Rest Timer** on.
3. Start a workout, complete a set to start a rest, then **lock the phone** and keep it locked past the deadline → the alert fires with sound.
4. Repeat with a Focus mode enabled (Do Not Disturb is enough) → the alert still sounds, because it is time-sensitive.
5. Repeat, but instead of locking, open the app switcher and hold it there past the deadline → the ping now shows and sounds instead of being swallowed.
6. Repeat with the app open on screen → unchanged: the in-app chime plays once, with no notification banner and no second ding.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

## Screenshots

No UI changes — the settings rows, their copy and the rest HUD are untouched.

## Notes for Reviewers

- **Verified on an iPhone by @CodeWithCJ**, running a build of this branch. The issue's own reproduction — app minimised, and phone locked — now plays the sound, which is its stated expected behaviour. The decisive run is the Focus one: with Do Not Disturb active the alert **played its sound**, which is precisely what the default `active` interruption level suppresses and what `timeSensitive` restores. A run without a Focus active would not have exposed that, since the ping's default sound played fine before this change in that case.
- **On vibration**, noted during that testing: this PR does not touch it and does not change it. The in-app haptic is gated on the app's own haptics preference and cannot fire while the phone is locked, since the app is not running then — any vibration in that state comes from the notification channel. Worth its own issue if it looks wrong, but it is independent of this diff.
- **Automated checks:** `pnpm run validate` is clean and the full mobile suite passes (424 suites / 6769 tests), with unit tests covering all three behaviours. The branch has `main` merged in and shows no conflicts.
- **Build impact:** the new entitlement is an App ID capability, so the first build carrying it needs credentials re-synced (managed EAS credentials regenerate the profile). Nothing else in the native config changed.
- **No translation churn.** The existing copy already matches the new behaviour — Workout Settings says the chime plays "while the app is open", Notification Settings says the alert fires "even in the background" — so no catalog was touched.
- **Considered and rejected:** shipping `rest-chime.wav` as the notification's own sound so the background cue matches the in-app chime. It needs the `expo-notifications` config plugin, which also injects an `aps-environment` push entitlement this app has no use for, and on Android 8+ the sound lives on the channel, which is immutable — changing it means a new channel id and losing every user's channel customisation. Worth revisiting separately if the default alert tone turns out to be too quiet.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rest-complete alerts now provide one clear cue: an in-app chime while the app is active, or a notification sound when it is in the background.
  * Background rest alerts are no longer cancelled prematurely, ensuring they arrive when the timer completes.
  * Rest notifications can now appear through Focus modes and Scheduled Summary on supported iOS devices.
  * Active-app rest alerts avoid unnecessary banners and notification-list entries.


zeusal:
I need answers to these questions https://github.com/CodeWithCJ/SparkyFitness/issues/2269#issuecomment-5562385818

zeusal:
I'm going to execute coderabbit

CodeWithCJ (approved):

PR #2419, opened 2026-09-10
fix(openfoodfacts): don't reuse metric variant values for household serving without serving_quantity

### LINKED ISSUE #2416: [Bug]: OpenFoodFacts household-serving variant reuses unscaled nutrient values
**Are you willing to submit a PR?** Yes, I'd like to submit a PR!

**What happened?**

For OpenFoodFacts products where the structured `serving_quantity` (grams) field is absent but the free-text `serving_size` field still parses as a household descriptor (e.g. `"2 tbsp (28 g)"`), `mapOpenFoodFactsProduct()` in `SparkyFitnessServer/integrations/openfoodfacts/openFoodFactsService.ts` builds a household-unit variant by copying the metric variant's nutrient values unchanged and only overwriting the serving label:

```js
const household = parseOffHouseholdServing(product.serving_size);
const householdVariant =
  household && !(household.size === metricVariant.serving_size && household.unit === metricVariant.serving_unit)
    ? { ...metricVariant, serving_size: household.size, serving_unit: household.unit, is_default: false }
    : null;
```

The comment above it assumes "it describes the SAME physical serving as the metric variant, so it reuses the exact same nutrient values — no rescaling." That's only true when `serving_quantity` was actually declared and used to scale `metricVariant`. When it's absent, `metricVariant` falls back to a 100g basis (unscaled), and the household variant then displays those same per-100g numbers under the household label — e.g. a product with 462 kcal/100g shows "2 tbsp = 462 kcal" instead of the correct ~60 kcal.

**Test coverage gap that let this ship:** `SparkyFitnessServer/tests/openFoodFactsService.test.ts`'s `describe('mapOpenFoodFactsProduct household serving variant')` block only has fixtures where `serving_quantity` IS set alongside a matching `serving_size` (the Milano cookies example: `serving_size: '2 cookies (28 g)'` with `serving_quantity: 28`) — the case where reuse is correct. None of the existing tests cover `serving_size` present with `serving_quantity` absent, which is the actual failure case.

**Steps to reproduce:** Any OFF product with a parseable household `serving_size` but no `serving_quantity` will produce a household variant whose nutrient values are the raw per-100g numbers, mislabeled.

**Expected behavior:** The household variant should not reuse `metricVariant`'s values unless `declaredServingQuantity` was actually present and used to scale it. Safest fix: skip emitting a household variant entirely when there's no verified gram amount backing it.

**How are you running SparkyFitness?** Docker Compose
**SparkyFitness version:** v1.6.5
**Which area is affected?** Server / API

georanma:
Fixes #2416.

OpenFoodFacts products with a free-text `serving_size` (e.g. "2 tbsp") but no structured `serving_quantity` were getting a household-unit variant that silently reused the metric variant's unscaled per-100g nutrient values under the household label — e.g. showing "2 tbsp = 462 kcal" for a product where 2 tbsp is actually ~60 kcal.

`parseOffHouseholdServing` now always returns a consistent `{ size: number | null; unit: string | null }` shape instead of `T | null`. The household-variant construction in `mapOpenFoodFactsProduct` additionally requires `declaredServingQuantity !== null` before reusing the metric variant's values — without a verified gram amount backing the label, no household variant is emitted at all.

**How to test:**
See the new test in `SparkyFitnessServer/tests/openFoodFactsService.test.ts` (`describe('mapOpenFoodFactsProduct household serving variant')`) — reproduces the bug with a synthetic product (parseable household `serving_size`, no `serving_quantity`) and asserts no household variant is emitted.

## PR Type
- [x] Issue (bug fix)

## Checklist

**All PRs:**

Draft PR — code and tests are ready and passing (typecheck/lint/full test file all green), opened as draft for easier review than a raw diff. Checkboxes left unchecked pending review.

### DIFF
--- a/SparkyFitnessServer/integrations/openfoodfacts/openFoodFactsService.ts
+++ b/SparkyFitnessServer/integrations/openfoodfacts/openFoodFactsService.ts
@@ -784,28 +784,47 @@ function deriveOffServingUnit(product: OffProduct): string {
 // duplicate the metric default (e.g. "28 g (28 g)").
 const METRIC_SERVING_UNITS = new Set(['g', 'ml', 'kg', 'l', 'oz']);
 
+// Result of parsing OFF's free-text serving_size for a household descriptor.
+// `size`/`unit` are both null when no such descriptor was found, rather than
+// returning null for the whole result — this keeps the "did we find a
+// household descriptor at all" question (this function's job) separate from
+// "is it safe to reuse the metric variant's values for it" (a question that
+// also depends on whether serving_quantity was declared, decided by the caller).
+interface HouseholdServing {
+  size: number | null;
+  unit: string | null;
+}
+
 // Extracts a household serving (e.g. "2 cookies") from OFF's free-text
 // serving_size string when it also states the equivalent metric weight/volume
 // in parentheses, e.g. "2 cookies (28 g)" or "1 cup (240 ml)". The parenthetical
-// is what confirms the household count maps to the same physical serving we
-// already computed from serving_quantity, so the household variant can safely
-// reuse the metric variant's nutrient values without any rescaling.
+// only confirms that OFF believes the household count maps to that metric
+// amount — it says nothing about whether serving_quantity was ever declared,
+// so callers must not assume the household count matches metricVariant's basis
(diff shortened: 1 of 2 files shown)


georanma (commented):
PR reviewed

PR #2429, opened 2026-09-11
feat(mobile): previous measurement hints and body fat calculator

### LINKED ISSUE #2428: [Feature]: Add previous measurement hints and body fat calculator to mobile check-in

The web Daily Check-In provides useful context when entering new body measurements: for standard measurements such as weight, height, neck, waist, hips and body fat, the most recent recorded value is displayed as a placeholder and can be adopted using the **Use last** action.

The mobile Measurements screen does not currently provide the same workflow. This makes recurring body measurements slower to enter and makes it harder to notice whether a new measurement differs significantly from the previous one.

The mobile implementation should also support **custom Daily measurement categories**, not only the built-in measurements.

There is also another web/mobile feature gap: the web check-in has a **Calculate** action for Body Fat %, while the mobile measurement form does not.

Bring the web measurement-entry conveniences to the mobile app.

#### 1. Previous-value hints for standard measurements

When the selected date does not already contain a value for a measurement:

* show the most recent value recorded on or before the selected date as a visually subdued / placeholder value;
* do **not** treat that placeholder as a new measurement automatically;
* provide a convenient way to adopt the previous value, equivalent to the web **Use last** action;
* once adopted, the value becomes a normal editable value and is saved normally.

This should apply where appropriate to standard measurements such as:

* Weight
* Height
* Neck
* Waist
* Hips
* Body Fat %
* BMR
* other compatible standard measurement fields

The exact list should follow the semantics already used by the web check-in.

#### 2. Support custom Daily measurement categories

The same previous-value behavior should also work for user-defined/custom **Daily** measurement categories.

For each custom category:

* find the most recent compatible value for that category on or before the selected date;
* display it as a hint when the selected date has no actual value;
* allow the user to adopt it;
* displaying the hint alone must never create or modify a measurement.

This should work for custom categories already supported by the mobile Daily Measurements form and should preserve their existing units and numeric/text handling.

#### 3. Body Fat % calculator on mobile

Add the Body Fat % calculation functionality available in the web check-in to the mobile measurement screen.

It should match the web behavior as closely as possible:

* respect the user's configured body-fat calculation algorithm;
* use the same calculation formulas/validation as the web implementation;
* use the required profile information such as gender/age where applicable;
* support the equivalent of the web **Use Recent** behavior, allowing recent Weight / Height / Waist / Neck / Hips measurements to be used when appropriate;
* write the calculated result into the Body Fat % input so that the user can review/edit it before saving;
* calculation itself must not save the check-in automatically.

### Expected behavior

A typical workflow should be:

1. Open Measurements on mobile for today.
2. Empty fields display the previous measurement in a subdued style.
3. Custom Daily categories behave the same way.
4. Tapping **Use last** (or the mobile equivalent) copies that value into the field.
5. Enter/update Neck, Waist, Height etc.
6. Tap **Calculate** next to Body Fat %.
7. Review the calculated percentage.
8. Save the measurements normally.

Existing values for the selected date should continue to behave exactly as they do today.

### Important data semantics

Previous values are **suggestions only**.

Simply opening the screen and saving unrelated fields must not copy previous measurements forward or create new custom measurement records.

This is important so that historical measurements remain real observations rather than silently generated duplicates.

Mobile app (Android / iOS), Measurements / Daily Check-In

Would significantly improve my workflow.

Yes. I can implement it and test an Android APK before marking the PR ready for review.

### Additional context

The web implementation can be used as the behavioral reference for both the latest-value hints and Body Fat % calculation.

I plan to test the mobile implementation using an APK built from my fork, provide before/after screenshots, and keep the resulting PR as a draft until the device testing is complete.

Dragonk:
## Description

The mobile Measurements screen lacked two conveniences the web Daily Check-In has had: previous-value hints for recurring body measurements (with a `Use last` action), and a Body Fat % `Calculate` action. This brings both to mobile, and extends the hints to custom Daily measurement categories — which the web check-in does not do.

*Previous-value hints (standard measurements)*
- The screen already loads the selected day's own row for the form. A second, separate query now loads the carry-forward row (`GET /api/measurements/check-in/latest-on-or-before-date`) purely for suggestions, so a suggestion is never confused with a real value.
- A hint is **display-only**: it lives in a derived object, never in `form` state. Rendering one therefore cannot submit it. `Use last` is the only path into state, and it writes through the normal field-update path, so the adopted value becomes dirty and behaves exactly like typed input.
- The selected day's own value always wins: a hint is offered only while the input is empty *and* the day holds no value for that field. Clearing a recorded value still clears it rather than resurrecting the suggestion.
- `steps` and `bmr` are deliberately same-day-only server-side, so they never generate a suggestion of their own.
- The suggestion query is a distinct cache entry from the day's own measurements, and saving refreshes it. Dirty-field tracking (already in place for refetches) is unchanged, so a background refresh cannot clobber input the user is editing.

*Previous values for custom Daily categories*
- The existing custom-entries list has no date filter at all, and `most-recent` only knows the fixed check-in columns, so "latest manual value per category on or before a day" was not obtainable. Per-category range calls would have been an N+1 for every category on screen.
- Added `GET /api/measurements/custom-entries/latest-manual-on-or-before-date`, which resolves every category in a single `DISTINCT ON` query. It is registered before `/custom-entries/:date` so the literal segment is not parsed as a date, and it validates the day as a real `YYYY-MM-DD` so a malformed value is a 400 rather than a 500 from the database.
- `source = 'manual'` is enforced in the query. A health-sync sample is historical context and must never be offered as something the user can adopt into a manual entry — so synced samples are not suggestions in the first place.
- Hints are keyed by category id, so a value cannot leak between categories.
- Older servers are handled explicitly. `/custom-entries/:date` also matches the literal bulk path and accepts it as a date, so an unimplemented endpoint fails as a 500 rather than a 404. When the bulk request fails, the client falls back to the long-standing list endpoint and narrows it by the same rules (manual sources only, nothing after the selected day, newest day then newest timestamp). That keeps the hints working against a server that predates this PR, in one request rather than one per category.
- The legacy fallback returns the hints its page proves rather than refusing a full page. The page is newest-first, so a category with an eligible entry inside it has its newest overall; a category whose only value fell outside the page simply gets no suggestion. A missing hint is acceptable, a wrong one is not, and rejecting the page would remove every hint on a sync-heavy account. The page is 2000 entries, and the fallback triggers on any HTTP failure except 401/403, since the status varies by server version.
 Only eligible Daily categories are considered, and the whole hint layer is gated on there being at least one eligible category.
- Boolean categories cannot show a placeholder, so they render the previous value as subdued text instead, with the same `Use last` control.

*Web check-in also gained custom-category previous values*
- The web Daily Check-In offered previous values for the standard fields only; custom Daily categories showed no context on what was last recorded. It now consumes the same endpoint and shows the same placeholder plus `Use last`, so both clients behave alike.
- The response shape moved to `@workspace/shared` because two hand-written interfaces would drift.

*Body Fat % calculator*
- The formulas moved to `@workspace/shared` so mobile and web run one implementation; `SparkyFitnessFrontend/src/services/bodyCompositionService.ts` remains the web import path and simply re-exports, so `BodyFatAlgorithm` and every call site are untouched. Existing web tests pass unchanged, which is the parity evidence.
- `Calculate` fills the Body Fat field and never saves — the user still presses Save.
- Validation is stricter than web in one direction only: web writes `0.00` when the Navy log argument is undefined, and would write a result above 100 that the field's own save validation rejects. Mobile reports a message instead of writing a value that cannot be kept. For every well-formed input the number is identical to web (covered by tests).

Linked Issue: Closes #2428

## How to Test

1. Check out this branch and run `pnpm install`, then from `SparkyFitnessMobile/`: `pnpm run validate` and `pnpm exec jest --watchman=false --runInBand`.
2. Open **Measurements** on mobile for a day with no recorded weight, on an account that has an older weight. The Weight field shows the older value as a subdued placeholder and a compact `Use last` appears next to the label.
3. Press `Use last` — the value becomes normal editable input. Edit an unrelated field and Save: the suggestion is *not* submitted unless adopted.
4. Pick a day that already has a weight: that value is editable as before, and no `Use last` is offered for it.
5. Add a custom **Daily** category with a value on an earlier day. Open a later day with no value for it: the previous value shows as a placeholder, and `Use last` adopts it.
6. Fill Height, Waist and Neck, then press `Calculate` next to Body Fat %. The field fills and nothing is saved.
7. Turn `Use recent` off and clear Height/Waist/Neck, then press `Calculate`: a validation message appears rather than a value.

## PR Type

- [ ] Issue (bug fix)
- [x] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

**New features only:**

  - No new tables, so no RLS change. The new endpoint reads `custom_measurements` under the existing user-scoped client and `checkin` permission middleware, exactly like the sibling custom-entries routes.

  - The After shot shows a custom Daily category carrying its previous value as a hint, and an actual value on the selected day taking precedence over it.

  - Tested on a physical Android device with `SparkyFitness-fd62a4b9-dev-release`, built from `fd62a4b9b6700ef637606a194ca4bd5e78cd4699` — the branch HEAD at testing time, containing every mobile change in this PR. Verified: previous-value hints for standard measurements and for custom Daily categories, `Use last` adopting them as normal input, and the Body Fat calculator filling the field without saving.
  - The branch has since advanced to `7d7bdaf9fb6183756e81d6e510585d214151726b`, whose only difference is one added entry in the web English translation catalog. No mobile source or test changed, so the tested behaviour carries over verbatim.

## Mobile testing status

- Automated tests: **complete** — mobile `pnpm run validate` passes (typecheck, lint, i18n generate-check + audit, Knip, native widget locales, Prettier), and the full mobile Jest suite is green (422 suites / 6724 tests). Server `pnpm run validate` + Vitest, and frontend `pnpm run validate` + Jest for the touched surfaces, also pass.
- APK build: **complete** (see the CI section below).
- Real-device validation: **complete** on `fd62a4b9b6700ef637606a194ca4bd5e78cd4699` (artifact `SparkyFitness-fd62a4b9-dev-release`), the branch HEAD at testing time. Current HEAD is `7d7bdaf9fb6183756e81d6e510585d214151726b`, which differs only by a web translation-catalog entry — no mobile code changed, so the verified behaviour is unchanged.

## Android test APK CI

Built from the fork's existing `build-test-apk.yml` (not modified by this PR):

| Field | Value |
|---|---|
| Workflow run | https://github.com/Dragonk/SparkyFitness/actions/runs/34623333491 |
| Repository | `Dragonk/SparkyFitness` |
| Branch | `feat/mobile-measurement-history-bf-calculator` |
| Tested commit SHA | `7d7bdaf9fb6183756e81d6e510585d214151726b` |
| App variant | `dev` |
| APK built successfully | **Yes** |
| Artifact name | `SparkyFitness-7d7bdaf9-dev-release` |

The workflow's `expected_sha` guard was set to the exact branch HEAD above, so the artifact corresponds to the recorded commit.

## Screenshots

<details>

### Before
Screenshot_2026-09-11-11-35-57-702_com SparkyApps SparkyFitnessMobile

### After
Screenshot_2026-09-11-11-35-42-389_org SparkyApps SparkyFitnessMobile1 dev
</details>

## Notes for Reviewers

- **Suggestion vs. data is the core of this change.** Hints never enter form state, so no code path can submit a value the user did not adopt. `Use last` writes through the existing field-update path, so an adopted value is indistinguishable from typed input.
- **One small server addition was necessary.** Mobile could not obtain "latest manual value per category on or before a day" without an N+1, so this adds one bulk endpoint rather than reusing a weaker query. It returns manual sources only by design.
- **Two deliberate divergences from web**, both narrow:
  - web passes a `0` through when the Navy formula's log argument is undefined; mobile reports it as uncomputable, because saving `0%` body fat is not useful and the field would reject a value above 100 anyway;
  - web has no custom-category hints at all, which is the point of extending this to mobile.
- **`steps` and `bmr` intentionally get no carried-forward suggestion.** The server treats them as same-day-only, and this change does not alter that.
- The formula move to `@workspace/shared` follows the repo's "rule of two": this was the second consumer, and a fork of the formulas would have drifted.
- The new mobile strings are English-only, per the localization contract; `pl`/`es` are Weblate-owned.

## Summary by CodeRabbit

- **New Features**
  - Added previous-value suggestions and “Use last” actions for standard and custom measurements.
  - Added body-fat calculations using U.S. Navy and BMI methods, with recent-value support and validation.
  - Added retrieval of the latest manual custom values by category.
  - Improved measurement form handling across metric and imperial units.
- **Bug Fixes**
  - Suggestions now refresh across dates after measurements are saved or deleted.
- **Tests**
  - Expanded coverage for suggestions, calculations, conversions, API behavior, validation, and fallback scenarios.


CodeWithCJ:
Could you look into the CI test failure.

CodeWithCJ (approved):

Issue #2002, opened 2026-08-01
[Feature]: Language select for search provider in iOS app


mariomaz87:
### Are you willing to submit a PR?

No, but I can help test

### What problem does this solve?

Hi, I’m Italian but I use most of my app in English. However I find difficult to identify exercises in English. Using the web app, if I select Italian as Sparky language, the interface is partly in Italian and I can search for exercises in Italian. In the iOS app, I can’t select the language and I have to search for exercises in English. 

### Describe the feature you'd like

I would like to select the language only for the search providers in the iOS app. 

### Which area does this relate to?

Mobile app (iOS)

### How important is this to you?

Would significantly improve my workflow

PR #2545, opened 2026-09-20
fix(mobile): send the active app language when searching and importing exercises

DUC750:
## Description

Exercise search in the mobile app returns nothing for non-English terms. The app never sends the `language` query parameter, so the server falls back to English and wger â€” which matches on `language__code` â€” finds no results. The same parameter is missing on import, so an exercise found in another language is stored under its English name.

`searchExternalExercises` and `importExercise` now resolve the language themselves through a new `getAppLanguageCode()` instead of taking it as an argument, so no call site can omit it. It returns the bare ISO 639-1 code (`de`, not `de-DE`) because the server passes the value straight to wger as `language__code`. The language is also part of the React Query key, so switching languages refetches instead of serving the previous language from cache.

Related: #2002 (same symptom reported for iOS)

## How to Test

Reproducible against the API without a device â€” the web app sends the parameter, the mobile app does not:

```
# nothing
GET /api/exercises/search-external?query=Beinpresse&providerType=wger&providerId=<id>

# three matches: Beinpresse, Beinpresse Eng, Beinpresse (weit)
GET /api/exercises/search-external?query=Beinpresse&providerType=wger&providerId=<id>&language=de
```

In the app: set the app language to a non-English one, search the external tab for a term in that language, and confirm results appear and the imported exercise keeps its localized name.

`pnpm jest __tests__/services/api/externalExerciseSearchApi.test.ts` â€” 12 passed (2 new: one asserts the search URL carries the active language, one asserts the wger import body does).
`pnpm jest __tests__/hooks/useExternalExerciseSearch.test.ts` â€” 5 passed.
`pnpm exec eslint` on the changed files â€” clean.

Rebased onto `main` after #2547; `pnpm run validate` passes end to end
(i18n:generate:check, typecheck, lint, i18n:audit, knip, native-locales:check,
format:check) and both affected suites are green â€” 17 tests.

I could not test on a device â€” happy to adjust if someone on iOS/Android spots something.

## PR Type

- [x] Issue (bug fix)
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation

## Checklist

**All PRs:**

  Verified on a physical device (Android 16, Xiaomi 13) using a development build of
  this branch: German search returns results, the imported exercise keeps its localized
  name, switching to English refetches instead of serving the cached German results, and
  previously saved exercises keep the language they were imported with.

---
## âœ… Required Checklist (restored automatically)

> has been added back below. **Please tick each box and do not delete this section** â€”
> no new commit is needed.


CodeWithCJ:
Is this tested via emulator or actual phone? all mobile changes needs to be tested to avoid surprise later

DUC750:
Not yet — honest answer: so far this is verified against a running server and the
unit tests only (`pnpm run validate` plus both affected suites pass on the current
`main` after the rebase).

I'm setting up a development build to run it on an actual phone, and I'll tick the
checkbox once that's done rather than before. What I'll check there: switch the app
language to a non-English one, search the external tab for a term in that language
and confirm results come back, then import one and confirm it keeps its localized
name instead of the English one.

Will report back here as soon as I've run it.

CodeWithCJ:
Marking this as draft. Let me know once its ready.

DUC750:
Tested on a physical device now — Android 16, Xiaomi 13, EAS development build of this branch.

- App language German: searching the external tab for "Beinpresse" returns results (previously empty).
- Importing one keeps the localized name instead of the English one.
- Switching to English and searching again returns the English results, so the language change takes effect rather than serving the previous language from cache.
- Exercises already saved keep the language they were imported with.

Server-side evidence from the access log, showing the parameter that was missing before:

```
GET /api/exercises/search-external?query=Leg+press&providerType=wger&providerId=...&page=1&pageSize=20&language=de  "okhttp/4.12.0"
```

One thing worth noting for anyone reviewing: with the app in German, searching "Leg press" does *not* return the leg press — it is translated to "Beinpresse" in wger's German catalogue. Untranslated exercises (JM Press, Arnold Press, …) do keep their English names there. That is expected behaviour of the data, not of this change.

Ticked the device checkbox accordingly.

CodeWithCJ (approved):

